PR #4, opened 2019-11-05
fix undesired update of dependency versions

neolit123:
The versions of these should be kept the same as in the
kubernetes/kubernetes repository at the time when we start
using the external validators there.

xref https://github.com/kubernetes/kubernetes/pull/84718#discussion_r342685017

/kind cleanup
/priority important-soon
/assign @yastij


neolit123:
> Not mergeable. Needs cncf-cla: yes

/check-cla

PR #5, opened 2019-11-07
fix misspell build for go 1.13

neolit123:
this was overlooked for go 1.12 and broke when we started using 1.13.

/kind bug
/priority important-soon
/assign @yastij

### DIFF
--- a/hack/verify-spelling.sh
+++ b/hack/verify-spelling.sh
@@ -38,7 +38,7 @@ URL="https://github.com/client9/misspell"
 echo "Cloning ${URL} in ${TMP_DIR}..."
 git clone --quiet --depth=1 "${URL}" "${TMP_DIR}"
 pushd "${TMP_DIR}" > /dev/null
-go mod init
+go mod init misspell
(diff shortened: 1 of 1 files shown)


hack/verify-spelling.sh
```diff
@@ -38,7 +38,7 @@ URL="https://github.com/client9/misspell"
 echo "Cloning ${URL} in ${TMP_DIR}..."
 git clone --quiet --depth=1 "${URL}" "${TMP_DIR}"
 pushd "${TMP_DIR}" > /dev/null
-go mod init
+go mod init misspell
```
yastij: are we aligning with the go version of k/k

neolit123: the version of go that is used in CI is the version in kubekins (test-infra) and that matches the version of k/k - 1.13.4.

yastij: sgtm

neolit123: /hold cancel

PR #11, opened 2019-12-15
Add preflight check for optional cgroups

odinuge:
pids and hugetlb are optional, ref.
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cm/cgroup_manager_linux.go#L325

### DIFF
--- a/validators/cgroup_validator.go
+++ b/validators/cgroup_validator.go
@@ -41,17 +41,22 @@ const (
 )
 
 // Validate is part of the system.Validator interface.
-func (c *CgroupsValidator) Validate(spec SysSpec) ([]error, []error) {
+func (c *CgroupsValidator) Validate(spec SysSpec) (warns, errs []error) {
 	subsystems, err := c.getCgroupSubsystems()
 	if err != nil {
 		return nil, []error{errors.Wrap(err, "failed to get cgroup subsystems")}
 	}
-	errs := c.validateCgroupSubsystems(spec.CgroupSpec.Required, subsystems, true)
-	warns := c.validateCgroupSubsystems(spec.CgroupSpec.Optional, subsystems, false)
-	return warns, errs
+	if missingRequired := c.validateCgroupSubsystems(spec.CgroupSpec.Required, subsystems, true); len(missingRequired) != 0 {
+		errs = []error{errors.Errorf("missing required cgroups: %s", strings.Join(missingRequired, " "))}
+	}
+	if missingOptional := c.validateCgroupSubsystems(spec.CgroupSpec.Optional, subsystems, false); len(missingOptional) != 0 {
+		warns = []error{errors.Errorf("missing optional cgroups: %s", strings.Join(missingOptional, " "))}
+	}
+	return
 }
 
-func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string, required bool) []error {
+// validateCgroupSubsystems returns a list with the missing cgroups in the cgroup
+func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string, required bool) []string {
 	var missing []string
 	for _, cgroup := range cgroups {
 		found := false
@@ -72,10 +77,7 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
 		}
 		missing = append(missing, cgroup)
 	}
-	if missing != nil {
-		return []error{errors.Errorf("missing cgroups: %s", strings.Join(missing, " "))}
-	}
-	return nil
+	return missing
 
 }
 
--- a/validators/cgroup_validator.go
+++ b/validators/cgroup_validator.go
@@ -46,15 +46,14 @@ func (c *CgroupsValidator) Validate(spec SysSpec) ([]error, []error) {
 	if err != nil {
 		return nil, []error{errors.Wrap(err, "failed to get cgroup subsystems")}
 	}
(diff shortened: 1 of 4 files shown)


neolit123:
@odinuge 
i'm ok with this change but i need sig-node folks to confirm they approve of this.
if you are not getting a review try asking in the #sig-node channel on k8s slack.

/hold
/approve
/sig node
/cc @Random-Liu

neolit123:
@kubernetes/sig-node-pr-reviews 
@kubernetes/sig-cluster-lifecycle-pr-reviews

validators/types_unix.go
```diff
@@ -56,7 +56,10 @@ var DefaultSysSpec = SysSpec{
 		},
 		Forbidden: []KernelConfig{},
 	},
-	Cgroups: []string{"cpu", "cpuacct", "cpuset", "devices", "freezer", "memory"},
+	CgroupSpec: CgroupSpec{
+		Required: []string{"cpu", "cpuacct", "cpuset", "devices", "freezer", "memory"},
+		Optional: []string{"hugetlb", "pids"},
```
RobertKrawitz: @dashpole IIRC there was a reason other than it not existing on all systems to not check for pids here, or am I remembering it incorrectly?

validators/cgroup_validator.go
```diff
@@ -46,15 +46,14 @@ func (c *CgroupsValidator) Validate(spec SysSpec) ([]error, []error) {
...
-	return nil, nil
+	errs := c.validateCgroupSubsystems(spec.CgroupSpec.Required, subsystems, true)
+	warns := c.validateCgroupSubsystems(spec.CgroupSpec.Optional, subsystems, false)
+	return warns, errs
 }
 
-func (c *CgroupsValidator) validateCgroupSubsystems(cgroupSpec, subsystems []string) error {
-	missing := []string{}
-	for _, cgroup := range cgroupSpec {
+func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string, required bool) []error {
```
RobertKrawitz: Style: might be simpler to merely report back the list of missing cgroups and let Validate above construct and report the appropriate errors and warnings.  Not going to insist on it if you prefer to do it this way, I just think it might mean less code change.

validators/types_unix.go
timothysc: Could you please document the reason why these are optional on certain versions?

validators/cgroup_validator.go
odinuge: Not sure which one i prefer to be honest. Updated to separate the concern now (as a new commit to make it easier to see the diff). Do you think the new one is better?

validators/types_unix.go
odinuge: > Could you please document the reason why these are optional on certain versions?

Sure! 

### hugetlb
The `hugetlb` cgroup is used for enforcing the limits on pods using [huge pages](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190129-hugepages.md#cgroup-enforcement). The current kubelet implementation (all versions with huge page support) just ignore it if the cgroup is missing, aka. not enforcing the limits. Some kernels are compiled without huge page support (resulting in the hugetlb cgroup to be missing). It is also possible to compile the kernel without the hugetlb cgroup but with huge page support, but I see no real reason why one should decide to to that).

The hugetlb cgroup is from 2012, so all "modern" kernels should support it if they support huge pages and cgroups. Setting it as "required" would then make sense, but since there might be reasons to compile the kernel without huge page support, i think keeping it optional is ok, or should it be required when huge pages are supported? If the hugetlb cgroup is not enabled on a node while huge pages are supported, the node will silently ignore it and _skip_ enforcing it (but that is another discussion).

Huge page support has been in `GA` since `v1.14`.


### pids
The `pids` cgroup is used for enforcing [pid limits](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190129-pid-limiting.md). There are two feature flags, `SupportPodPidsLimit` & `SupportNodePidsLimit`. Both are in beta (from `v1.14` and `v1.15`), and therefore enabled by default. The current implementation will require the `pids` cgroup as long as one of them is enabled, and since both are enabled by default, the cgroup is essentially required.

The `pids` cgroup was introduced in 2015, so it is a bit newer than `hugetlb`. @RobertKrawitz and @dashpole probably have some more insights.

dashpole: Yeah, unless the validator has access to feature gates, it seems reasonable to mark pids as optional.

validators/cgroup_validator.go
RobertKrawitz: Sorry, didn't see this until just now.  I prefer this, but not strongly enough to want to insist on it.

validators/types_unix.go
RobertKrawitz: @odinuge could you put a brief summary of this in the code so that when someone looks at it they'll understand the rationale?

neolit123: > Could you please document the reason why these are optional on certain versions?

i think having some comments in the code make sense too.
possibly not the full explanation, but a short summary @odinuge

timothysc: +1 to code comment.

odinuge:
Updated the PR with comments now :smile:

odinuge:
ping @neolit123

neolit123:
based on the discussion and review.
/lgtm
/hold cancel

thanks for the updates.
to enable usage of this in k/k a new TAG for this repository and another PR in k/k is required.

PR #12, opened 2020-02-06
Support for cgroupv2

KentaTada:
Preflight check will fail when clusters use cgroupv2 in future because /proc/cgroups is meaningless for v2.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,4 +6,5 @@ require (
 	github.com/blang/semver v3.5.0+incompatible
 	github.com/pkg/errors v0.8.0
 	github.com/stretchr/testify v1.3.0
+	golang.org/x/sys v0.0.0-20191022100944-742c48ecaeb7
 )
--- a/validators/cgroup_validator.go
+++ b/validators/cgroup_validator.go
@@ -18,10 +18,13 @@ package system
 
 import (
 	"bufio"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
+	"golang.org/x/sys/unix"
 )
 
 var _ Validator = &CgroupsValidator{}
@@ -38,18 +41,42 @@ func (c *CgroupsValidator) Name() string {
 
 const (
 	cgroupsConfigPrefix = "CGROUPS_"
+	unifiedMountpoint   = "/sys/fs/cgroup"
 )
 
 // Validate is part of the system.Validator interface.
 func (c *CgroupsValidator) Validate(spec SysSpec) (warns, errs []error) {
-	subsystems, err := c.getCgroupSubsystems()
-	if err != nil {
-		return nil, []error{errors.Wrap(err, "failed to get cgroup subsystems")}
+	// Get the subsystems from /sys/fs/cgroup/cgroup.controllers when cgroup v2 is used.
+	// /proc/cgroups is meaningless for v2
+	// https://github.com/torvalds/linux/blob/v5.3/Documentation/admin-guide/cgroup-v2.rst#deprecated-v1-core-features
+	var st unix.Statfs_t
+	var err error
+	if err := unix.Statfs(unifiedMountpoint, &st); err != nil {
+		return nil, []error{errors.Wrap(err, "cannot statfs the cgroupv2 root")}
 	}
-	if missingRequired := c.validateCgroupSubsystems(spec.CgroupSpec.Required, subsystems, true); len(missingRequired) != 0 {
+	var requiredCgroupSpec []string
+	var optionalCgroupSpec []string
+	var subsystems []string
+	if st.Type == unix.CGROUP2_SUPER_MAGIC {
+		subsystems, err = c.getCgroupV2Subsystems()
+		if err != nil {
+			return nil, []error{errors.Wrap(err, "failed to get cgroup v2 subsystems")}
+		}
+		requiredCgroupSpec = spec.CgroupV2Spec.Required
+		optionalCgroupSpec = spec.CgroupV2Spec.Optional
+	} else {
+		subsystems, err = c.getCgroupV1Subsystems()
+		if err != nil {
+			return nil, []error{errors.Wrap(err, "failed to get cgroup v1 subsystems")}
+		}
+		requiredCgroupSpec = spec.CgroupSpec.Required
+		optionalCgroupSpec = spec.CgroupSpec.Optional
+	}
+
+	if missingRequired := c.validateCgroupSubsystems(requiredCgroupSpec, subsystems, true); len(missingRequired) != 0 {
 		errs = []error{errors.Errorf("missing required cgroups: %s", strings.Join(missingRequired, " "))}
 	}
-	if missingOptional := c.validateCgroupSubsystems(spec.CgroupSpec.Optional, subsystems, false); len(missingOptional) != 0 {
+	if missingOptional := c.validateCgroupSubsystems(optionalCgroupSpec, subsystems, false); len(missingOptional) != 0 {
 		warns = []error{errors.Errorf("missing optional cgroups: %s", strings.Join(missingOptional, " "))}
 	}
 	return
@@ -81,7 +108,8 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
 
 }
 
-func (c *CgroupsValidator) getCgroupSubsystems() ([]string, error) {
+func (c *CgroupsValidator) getCgroupV1Subsystems() ([]string, error) {
+	// Get the subsystems from /proc/cgroups when cgroup v1 is used.
 	f, err := os.Open("/proc/cgroups")
 	if err != nil {
 		return nil, err
@@ -102,5 +130,22 @@ func (c *CgroupsValidator) getCgroupSubsystems() ([]string, error) {
 			}
 		}
 	}
+	return subsystems, nil
+}
+
+func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
+	// Some controllers are implicitly enabled by the kernel.
+	// Those controllers do not appear in /sys/fs/cgroup/cgroup.controllers.
+	// https://github.com/torvalds/linux/blob/v5.3/kernel/cgroup/cgroup.c#L433-L434
+	// We assume these are always available, as it is hard to detect availability.
+	// So, we hardcode the following as "pseudo" controllers.
+	// - devices: implemented in kernel 4.15
+	// - freezer: implemented in kernel 5.2
+	pseudo := []string{"devices", "freezer"}
+	data, err := ioutil.ReadFile(filepath.Join(unifiedMountpoint, "cgroup.controllers"))
+	if err != nil {
+		return nil, err
+	}
+	subsystems := append(pseudo, strings.Fields(string(data))...)
 	return subsystems, nil
 }
--- a/validators/cgroup_validator_test.go
+++ b/validators/cgroup_validator_test.go
@@ -23,6 +23,9 @@ import (
 )
 
 func TestValidateCgroupSubsystem(t *testing.T) {
+	// hardcoded cgroup v2 subsystems
+	pseudoSubsystems := []string{"devices", "freezer"}
+
 	v := &CgroupsValidator{
 		Reporter: DefaultReporter,
 	}
@@ -53,6 +56,30 @@ func TestValidateCgroupSubsystem(t *testing.T) {
 		"subsystems the same with spec should not report missing": {
 			cgroupSpec: []string{"system1"},
 			subsystems: []string{"system1", "system2"},
+			required:   false,
+			missing:    nil,
+		},
+		"missing required cgroup subsystem when pseudo hardcoded subsystems are set": {
+			cgroupSpec: []string{"system1", "devices", "freezer"},
+			subsystems: append(pseudoSubsystems),
+			required:   true,
+			missing:    []string{"system1"},
+		},
+		"missing optional cgroup subsystem when pseudo hardcoded subsystems are set": {
+			cgroupSpec: []string{"system1", "devices", "freezer"},
+			subsystems: append(pseudoSubsystems),
+			required:   false,
+			missing:    []string{"system1"},
+		},
+		"extra cgroup subsystems when pseudo hardcoded subsystems are set": {
+			cgroupSpec: []string{"system1", "devices", "freezer"},
+			subsystems: append(pseudoSubsystems, "system1", "system2"),
+			required:   true,
+			missing:    nil,
+		},
+		"matching list of cgroup subsystems including pseudo hardcoded subsystems": {
+			cgroupSpec: []string{"system1", "devices", "freezer"},
+			subsystems: append(pseudoSubsystems, "system1"),
 			required:   false,
 			missing:    nil,
 		},
--- a/go.sum
+++ b/go.sum
@@ -9,3 +9,5 @@ github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZN
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/testify v1.3.0 h1:TivCn/peBQ7UY8ooIcPgZFpTNSz0Q2U6UrFlUfqbe0Q=
 github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UVUgZn+9EI=
+golang.org/x/sys v0.0.0-20191022100944-742c48ecaeb7 h1:HmbHVPwrPEKPGLAcHSrMe6+hqSUlvZU0rab6x5EXfGU=
+golang.org/x/sys v0.0.0-20191022100944-742c48ecaeb7/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
--- a/validators/types.go
+++ b/validators/types.go
@@ -120,8 +120,10 @@ type SysSpec struct {
 	OS string `json:"os,omitempty"`
 	// KernelConfig defines the spec for kernel.
 	KernelSpec KernelSpec `json:"kernelSpec,omitempty"`
-	// Cgroups is the required cgroups.
+	// Cgroups is the required cgroups v1.
 	CgroupSpec CgroupSpec `json:"cgroupSpec,omitempty"`
+	// Cgroups is the required cgroups v2.
+	CgroupV2Spec CgroupSpec `json:"cgroupV2Spec,omitempty"`
 	// RuntimeSpec defines the spec for runtime.
 	RuntimeSpec RuntimeSpec `json:"runtimeSpec,omitempty"`
 	// PackageSpec defines the required packages and their versions.
--- a/validators/types_unix.go
+++ b/validators/types_unix.go
@@ -67,6 +67,10 @@ var DefaultSysSpec = SysSpec{
 			"pids",
 		},
 	},
+	CgroupV2Spec: CgroupSpec{
+		Required: []string{"cpu", "cpuset", "devices", "freezer", "memory"},
+		Optional: []string{"hugetlb", "pids"},
+	},
 	RuntimeSpec: RuntimeSpec{
 		DockerSpec: &DockerSpec{
 			Version:     []string{`1\.1[1-3]\..*`, `17\.0[3,6,9]\..*`, `18\.0[6,9]\..*`, `19\.03\..*`},


neolit123:
@kubernetes/sig-node-pr-reviews 
/ok-to-test
/assign

neolit123:
@kubernetes/sig-cluster-lifecycle-pr-reviews

validators/cgroup_validator.go
```diff
@@ -38,8 +41,26 @@ func (c *CgroupsValidator) Name() string {
...
+// IsCgroup2UnifiedMode returns whether we are running in cgroup v2 unified mode.
+func IsCgroup2UnifiedMode() bool {
+	isUnifiedOnce.Do(func() {
+		var st unix.Statfs_t
+		if err := unix.Statfs(unifiedMountpoint, &st); err != nil {
+			panic("cannot statfs cgroup root")
+		}
+		isUnified = st.Type == unix.CGROUP2_SUPER_MAGIC
+	})
+	return isUnified
```
neolit123: given this function is only used in getCgroupSubsystems() i would unroll it there and remove the sync.Once usage and the global variables on line 47.

```
var st unix.Statfs_t
if err := unix.Statfs(unifiedMountpoint, &st); err != nil {
	panic("cannot statfs cgroup root")
}
if st.Type == unix.CGROUP2_SUPER_MAGIC {
	// handle cgroupv2

}
```

validators/cgroup_validator.go
```diff
@@ -102,5 +138,8 @@ func (c *CgroupsValidator) getCgroupSubsystems() ([]string, error) {
 			}
 		}
 	}
+	if err := s.Err(); err != nil {
+		return nil, err
+	}
```
neolit123: could you explain why this is needed?
this is already handled on line 130.

validators/cgroup_validator.go
```diff
@@ -82,6 +103,21 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
 }
 
 func (c *CgroupsValidator) getCgroupSubsystems() ([]string, error) {
+	// /proc/cgroups is meaningless for v2
+	// https://github.com/torvalds/linux/blob/v5.3/Documentation/admin-guide/cgroup-v2.rst#deprecated-v1-core-features
+	if IsCgroup2UnifiedMode() {
+		// "pseudo" controllers do not appear in /sys/fs/cgroup/cgroup.controllers.
+		// - devices: implemented in kernel 4.15
+		// - freezer: implemented in kernel 5.2
```
neolit123: i think having this list of the two controllers needs a separate reference link explaining why they are picked.

validators/cgroup_validator.go
```diff
@@ -82,6 +103,21 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
...
+		// We assume these are always available, as it is hard to detect availability.
```
neolit123: is this because they are already mounted / bound to the v1 hierarchy?
https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#mounting

validators/cgroup_validator.go
```diff
@@ -82,6 +103,21 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
...
+		pseudo := []string{"devices", "freezer"}
```
neolit123: is this because they are already mounted / bound to the v1 hierarchy?
https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html#mounting

here is a freezer for v2, https://lwn.net/Articles/772377/
with this hardcoded assumption, can the final list of controllers end up having two "freezers"?

validators/cgroup_validator.go
```diff
@@ -82,6 +103,21 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
...
+		data, err := ioutil.ReadFile("/sys/fs/cgroup/cgroup.controllers")
+		if err != nil {
+			return nil, err
+		}
+		subsystems := append(pseudo, strings.Fields(string(data))...)
```
neolit123: i guess we'd need cgroup_validator_test.go unit tests for this.

KentaTada:
Thank you. I'm sorry I mistakenly pushed my old commit. I re-pushed it.
But I need to confirm remained comments. I'll handle it.

validators/cgroup_validator.go
KentaTada: It's my mistake from the old commit. I removed this code.

validators/cgroup_validator.go
neolit123: it might be better to not panic here and instead return this as a:
`return []string, errors.Wrapf(err, "cannot statfs the cgroupv2 root")`

the `err` in this case should already include the path

validators/cgroup_validator.go
KentaTada: Recent kernel does not show all controllers in /sys/fs/cgroup/cgroup.controllers. This line
https://github.com/torvalds/linux/blob/v5.3/kernel/cgroup/cgroup.c#L433-L434 actually decides what cgroup subsystems are masked. So "devices" and "freezer" are not visible.

validators/cgroup_validator.go
KentaTada: "freezers" is not visible in /sys/fs/cgroup/cgroup.controllers as described above. "freezers" don' t have dfl_cftypes and the bit of cgrp_dfl_implicit_ss_mask of "freezers" is set. So the final list does not have two "freezers".

KentaTada: In addition to that, "freezers" may set the flag of cgroup_subsys->implicit_on_dfl in future https://lore.kernel.org/patchwork/patch/651321/

validators/cgroup_validator.go
KentaTada: Thank you. I modified this implementation.

validators/cgroup_validator.go
KentaTada: @neolit123 That is reasonable but it is difficult to test the current implementation of getCgroupSubsystems(). Should we refactor it or do you have any ideas?

validators/cgroup_validator.go
neolit123: i think reference links should be added above these hardcoded controllers, explaining the reasoning.

validators/cgroup_validator.go
neolit123: current unit tests are only for validateCgroupSubsystems(). the hardcoded / preudo controllers should be unit tested at least.

validators/cgroup_validator.go
KentaTada: I added the reference link and explanation.

validators/cgroup_validator.go
KentaTada: I added unit tests for validateCgroupSubsystems() to test when hardcoded subsystems values("devices" and "freezer") are set.

KentaTada:
@neolit123 PTAL
Thank you!

go.mod
```diff
@@ -6,4 +6,5 @@ require (
 	github.com/blang/semver v3.5.0+incompatible
 	github.com/pkg/errors v0.8.0
 	github.com/stretchr/testify v1.3.0
+	golang.org/x/sys v0.0.0-20190812073006-9eafafc0a87e
```
neolit123: can we use the same version as k/k?
https://github.com/kubernetes/kubernetes/blob/master/go.mod#L127

`v0.0.0-20191022100944-742c48ecaeb7`

validators/cgroup_validator.go
```diff
@@ -82,6 +85,31 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
...
 func (c *CgroupsValidator) getCgroupSubsystems() ([]string, error) {
+	// Get the subsystems from /sys/fs/cgroup/cgroup.controllers when cgroup v2 is used.
+	// /proc/cgroups is meaningless for v2
+	// https://github.com/torvalds/linux/blob/v5.3/Documentation/admin-guide/cgroup-v2.rst#deprecated-v1-core-features
+	var st unix.Statfs_t
+	if err := unix.Statfs(unifiedMountpoint, &st); err != nil {
+		return []string{}, errors.Wrap(err, "cannot statfs the cgroupv2 root")
+	}
+	if st.Type == unix.CGROUP2_SUPER_MAGIC {
+		// Some controllers are implicitly enabled by kernel.
```
neolit123: `by the kernel`

validators/cgroup_validator.go
```diff
@@ -82,6 +85,31 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
...
+		// Those controllers do not appear in /sys/fs/cgroup/cgroup.controllers.
+		// https://github.com/torvalds/linux/blob/v5.3/kernel/cgroup/cgroup.c#L433-L434
+		// We assume these are always available, as it is hard to detect availability.
+		// So, we hardcoded "pseudo" controllers.
```
neolit123: > So, we hardcode the following as "pseudo" controllers.

validators/cgroup_validator_test.go
```diff
@@ -56,6 +59,30 @@ func TestValidateCgroupSubsystem(t *testing.T) {
 			required:   false,
 			missing:    nil,
 		},
+		"missing required group subsystem when pseudo hardcoded subsystems are set should report missing": {
```
neolit123: > missing required cgroup subsystem when pseudo hardcoded subsystems are set

validators/cgroup_validator_test.go
```diff
@@ -56,6 +59,30 @@ func TestValidateCgroupSubsystem(t *testing.T) {
...
+			cgroupSpec: []string{"system1", "devices", "freezer"},
+			subsystems: append(pseudoSubsystems),
+			required:   true,
+			missing:    []string{"system1"},
+		},
+		"missing optional cgroup subsystem when pseudo hardcoded subsystems are set should report missing": {
```
neolit123: > missing optional cgroup subsystem when pseudo hardcoded subsystems are set

validators/cgroup_validator_test.go
```diff
@@ -56,6 +59,30 @@ func TestValidateCgroupSubsystem(t *testing.T) {
...
+			cgroupSpec: []string{"system1", "devices", "freezer"},
+			subsystems: append(pseudoSubsystems),
+			required:   false,
+			missing:    []string{"system1"},
+		},
+		"extra cgroup subsystems when pseudo hardcoded subsystems are set should not report missing": {
```
neolit123: > extra cgroup subsystems when pseudo hardcoded subsystems are set

validators/cgroup_validator_test.go
```diff
@@ -56,6 +59,30 @@ func TestValidateCgroupSubsystem(t *testing.T) {
...
+			cgroupSpec: []string{"system1", "devices", "freezer"},
+			subsystems: append(pseudoSubsystems, "system1", "system2"),
+			required:   true,
+			missing:    nil,
+		},
+		"subsystems the same with spec when pseudo hardcoded subsystems are set should not report missing": {
```
neolit123: > matching list of cgroup subsystems including pseudo hardcoded subsystems

neolit123:
thanks for the update @KentaTada 
i will ping the owners of the kubelet (SIG-Node) for a review in their slack channel.

odinuge:
The code looks good to me, but not sure if we should do this yet (altho this is quite generic). cgroupv2 is still (at least) a few releases away in k/k, and the support is tracked here: https://github.com/kubernetes/enhancements/pull/1370.

k/k do currently _not_ support nodes with cgroupv2 enabled by default (mounted on that mount point as this PR looks for) as is now, so this will in practice make nodes pass validation, without being supported. kubelet still require all the cgroupv1 cgroups to be enabled.

So, ill add a hold.
/hold

Another thought is that we maybe should start mocking the fs in the test (by setting the base directory) to make it simpler to do e2e-like testing of this, since

neolit123:
i'm personally fine with merging this even if the actual support is anticipated in the near future.

odinuge:
> i'm personally fine with merging this even if the actual support is anticipated in the near future.

Not sure when cgroupv2 will be fully supported by k/k tho. Definitely depending on the definition of `near future`, and it will certainly not be ready for 1.18.

The only issue is when someone running newest distros like the newest version of Fedora, and use kubeadm to bootstrap. All preflight checks will report ok, but kubelet will keep crashing due to missing cgroups. Other than that, it will make no hurt I guess.

So ill defer this to @derekwaynecarr.

neolit123:
> The only issue is when someone running newest distros like the newest version of Fedora, and use kubeadm to bootstrap. All preflight checks will report ok, but kubelet will keep crashing due to missing cgroups. Other than that, it will make no hurt I guess.

k/k can remain importing the old version of the system-validators repository, but this will prevent merging other fixes in the meantime, so it seems better to hold because of that.

a problem is contributor bandwith too. if in N release @KentaTada is not available to update the PR, someone has to sign to take over.

KentaTada:
Thank you for your review!

This is just FYI but I want to share our background to support cgroupv2 for Kubernetes.
As Sony, we are building on-premise Kubernetes clusters for our rootless embedded container system and our system needs cgroupv2. So, I found this issue when I integrated Kubernetes with our system.

I think that it may be helpful for people like us as same as Fedora users if cgroupv2 support is prepared in mainline.

neolit123:
> So, I found this issue when I integrated Kubernetes with our system.

are you using kubeadm which fails a cgroup preflight check?

KentaTada:
> > So, I found this issue when I integrated Kubernetes with our system.
> 
> are you using kubeadm which fails a cgroup preflight check?

No. I'm sorry to confuse you but I found this issue when I checked source code before the dynamic test.

neolit123:
> No. I'm sorry to confuse you but I found this issue when I checked source code before the dynamic test.

do you mean that you are not seeing any problems yet when these system-validators are used on your Nodes - i.e. you saw the verification is missing, but there is no real issue yet?

KentaTada:
> > No. I'm sorry to confuse you but I found this issue when I checked source code before the dynamic test.
> 
> do you mean that you are not seeing any problems yet when these system-validators are used on your Nodes - i.e. you saw the verification is missing in the source code, but there is no real issue yet?

I have not experienced the real issue yet.

neolit123:
> I have not experienced the real issue yet.

ok, thanks.
let's see if other folks from SIG Node have an opinion here. it seems better to hold the PR until cgroupv2 support in k8s is ready.

giuseppe:
> let's see if other folks from SIG Node have an opinion here. it seems better to hold the PR until cgroupv2 support in k8s is ready.

I think these two could happen at the same time.  I already have a PR for supporting cgroup v2 in the Kubelet, but I am still waiting for the KEP to be accepted.

dchen1107:
Thanks for sending the pr. cc @yguo0905 

Can you hold that until cgroup v2 is properly supported by node (kubelet, container runtime, cadvisor, etc.)?

go.mod
KentaTada: I'll update it when cgroupv2 is actually merged.

validators/cgroup_validator.go
KentaTada: fix it. Thank you.

validators/cgroup_validator.go
KentaTada: fix it. Thank you.

validators/cgroup_validator_test.go
KentaTada: fix it. Thank you.

validators/cgroup_validator_test.go
KentaTada: fix it. Thank you.

validators/cgroup_validator_test.go
KentaTada: fix it. Thank you.

validators/cgroup_validator_test.go
KentaTada: fix it. Thank you.

giuseppe:
> Thanks for sending the pr. cc @yguo0905
> 
> Can you hold that until cgroup v2 is properly supported by node (kubelet, container runtime, cadvisor, etc.)?

@dchen1107 the cgroup v2 PR for Kubelet got merged and Linux 5.6 was released (this is needed for the hugetlb controller).  I'd expect everything to work on cgroup v2 at this point

go.mod
giuseppe: @KentaTada cgroup v2 support for the Kubelet was merged, I think this PR can move forward

KentaTada: @giuseppe Thank you for your info. I did not know when was suitable for merge because  https://github.com/kubernetes/kubernetes/pull/89567 was still ongoing.

neolit123:
i see that https://github.com/kubernetes/kubernetes/pull/85218 merged.

looking for LGTM by sig-node so that we can merge this PR, create a new release in this repository and vendor the new release in kubernetes/kubernetes.

validators/cgroup_validator.go
```diff
@@ -82,6 +85,31 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
...
 
 func (c *CgroupsValidator) getCgroupSubsystems() ([]string, error) {
+	// Get the subsystems from /sys/fs/cgroup/cgroup.controllers when cgroup v2 is used.
+	// /proc/cgroups is meaningless for v2
+	// https://github.com/torvalds/linux/blob/v5.3/Documentation/admin-guide/cgroup-v2.rst#deprecated-v1-core-features
+	var st unix.Statfs_t
+	if err := unix.Statfs(unifiedMountpoint, &st); err != nil {
+		return []string{}, errors.Wrap(err, "cannot statfs the cgroupv2 root")
+	}
+	if st.Type == unix.CGROUP2_SUPER_MAGIC {
```
odinuge, line 95: Maybe split these cases to make them easier to test?

validators/cgroup_validator.go
```diff
@@ -82,6 +85,31 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
...
+		// Some controllers are implicitly enabled by the kernel.
+		// Those controllers do not appear in /sys/fs/cgroup/cgroup.controllers.
+		// https://github.com/torvalds/linux/blob/v5.3/kernel/cgroup/cgroup.c#L433-L434
+		// We assume these are always available, as it is hard to detect availability.
+		// So, we hardcode the following as "pseudo" controllers.
+		// - devices: implemented in kernel 4.15
+		// - freezer: implemented in kernel 5.2
+		pseudo := []string{"devices", "freezer"}
+		data, err := ioutil.ReadFile("/sys/fs/cgroup/cgroup.controllers")
```
odinuge, line 104: nit: This can reuse the `unifiedMountpoint` constant.

KentaTada: I reused unifiedMountpoint.  Thanks!

KentaTada:
@odinuge 
My understanding is that we need to modify below to test both cgroup v1 and v2.

- Add the new DefaultSysSpec for cgroup v2
https://github.com/kubernetes/system-validators/blob/master/validators/types_unix.go
- Modify kubeadm to use cgroup v1 or v2.
https://github.com/kubernetes/kubernetes/blob/v1.19.0-alpha.1/cmd/kubeadm/app/preflight/checks.go#L546
- Modify e2e test to use cgroup v1 or v2.
https://github.com/kubernetes/kubernetes/blob/v1.19.0-alpha.1/test/e2e_node/e2e_node_suite_test.go#L133

In addition to that, should we split getCgroupSubsystems like getCgroupv1Subsystems and getCgroupv2Subsystems to make test easy?

If my understanding is correct, I'll prepare for it.

odinuge:
> @odinuge
> My understanding is that we need to modify below to test both cgroup v1 and v2.
> 
>    Add the new DefaultSysSpec for cgroup v2
    https://github.com/kubernetes/system-validators/blob/master/validators/types_unix.go
    Modify kubeadm to use cgroup v1 or v2.
    https://github.com/kubernetes/kubernetes/blob/v1.19.0-alpha.1/cmd/kubeadm/app/preflight/checks.go#L546
    Modify e2e test to use cgroup v1 or v2.
    https://github.com/kubernetes/kubernetes/blob/v1.19.0-alpha.1/test/e2e_node/e2e_node_suite_test.go#L133
> 
> In addition to that, should we split getCgroupSubsystems like getCgroupv1Subsystems and getCgroupv2Subsystems to make test easy?
> 
> If my understanding is correct, I'll prepare for it.

Yes, that sounds like a plan! I guess @neolit123 may have some more input.

There will also be some various differences when it comes to the kernel config flags too. Here are some of my quick observations:
- `CGROUP_DEVICE` - Still required in for cgroupv2 only configs
- `CGROUP_CPUACCT` - Not required (as far as i know, not 100% sure tho, so just keep as required for now) for cgroupv2 only systems
- `CGROUP_SCHED` - Not required (as far as i know, not 100% sure tho, so just keep as required for now) for cgroupv2 only systems
- `CONFIG_CGROUP_BPF` - Required for cgroupv2 (for controlling devices)

neolit123:
> Add the new DefaultSysSpec for cgroup v2
https://github.com/kubernetes/system-validators/blob/master/validators/types_unix.go

this is not ideal. instead of adding a new `DefaultSysSpec` structure there should be a new `CgroupV2Spec` field here:
https://github.com/kubernetes/system-validators/blob/f434adad8f2aeb4bee4a2f29fdfeefbac2d05bd9/validators/types_unix.go#L69

> Modify kubeadm to use cgroup v1 or v2.
> Modify e2e test to use cgroup v1 or v2.

not needed.

KentaTada:
@odinuge @neolit123 Thank you for your help :satisfied:
I think we can prepare for kernel config flags each cgroup version if we add a new DefaultSysSpec. But it is difficult to maintain.
As of now, I just only add a new CgroupV2Spec field and don't modify the kubeadm and the e2e test.

validators/cgroup_validator.go
```diff
@@ -81,7 +108,25 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
 
 }
 
-func (c *CgroupsValidator) getCgroupSubsystems() ([]string, error) {
+func (c *CgroupsValidator) getCgroupv2Subsystems() ([]string, error) {
```
neolit123, line 111: getCgroupv2Subsystems -> getCgroupV2Subsystems

validators/cgroup_validator.go
```diff
@@ -81,7 +108,25 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
...
+	pseudo := []string{"devices", "freezer"}
+	data, err := ioutil.ReadFile(filepath.Join(unifiedMountpoint, "cgroup.controllers"))
+	if err != nil {
+		return nil, err
+	}
+	subsystems := append(pseudo, strings.Fields(string(data))...)
+	return subsystems, nil
+}
+
+func (c *CgroupsValidator) getCgroupv1Subsystems() ([]string, error) {
```
neolit123, line 128: getCgroupv1Subsystems -> getCgroupV1Subsystems

validators/cgroup_validator.go
```diff
@@ -38,18 +41,42 @@ func (c *CgroupsValidator) Name() string {
...
+		subsystems, err = c.getCgroupv2Subsystems()
+		if err != nil {
+			return nil, []error{errors.Wrap(err, "failed to get cgroup v2 subsystems")}
+		}
+		requiredCgroupSpec = spec.CgroupV2Spec.Required
+		optionalCgroupSpec = spec.CgroupV2Spec.Optional
+	} else {
+		subsystems, err = c.getCgroupv1Subsystems()
+		if err != nil {
+			return nil, []error{errors.Wrap(err, "failed to get cgroup v2 subsystems")}
```
neolit123, line 70: v2 subsystems -> v1 subsystems

validators/cgroup_validator.go
neolit123: please move the v2 function bellow the v1 function.

KentaTada: Fixed it

validators/cgroup_validator.go
KentaTada: Fixed it

validators/cgroup_validator.go
KentaTada: Fixed it

validators/cgroup_validator.go
KentaTada: Fixed it

KentaTada:
@neolit123 I modified cgroup_validator.go

odinuge:
This looks good to me.

The only thing is that the kernel config `CONFIG_CGROUP_BPF=y` is required when using cgroupv2 (used for controlling device access). We cannot set it required for all, since a lot of older kernels have been compiled without it. Or should we just don't care about it?

odinuge:
Other than `CONFIG_CGROUP_BPF=y`, this looks good to me now. I think we should get this into k/k as quick as possible, so we can take that flag later. Noone will compile their kernel without bpf for cgroups and use cgroup v2, so it shouldn't be a big problem yet.

/lgtm
/hold cancel

ping @neolit123

neolit123:
> Other than CONFIG_CGROUP_BPF=y, this looks good to me now. I think we should get this into k/k as quick as possible, so we can take that flag later. Noone will compile their kernel without bpf for cgroups and use cgroup v2, so it shouldn't be a big problem yet.

agreed, popular kernels should have CGROUP_BPF.
if we face a single case without it we can add the extra validation for v2.
although, hoping that the person doing triage on such a bug report will know what to ask for.

/approve

i will create a new release and send the PR for k/k to update the vendored validators.

PR #14, opened 2020-04-06
properly perform build verification and skip some tests on non-Linux

neolit123:
fix issues that blocked this PR from merging in k/k:
https://github.com/kubernetes/kubernetes/pull/89901

contains 4 commits:

```
cgroup_validator: don't validate cgroups on non-Linux
package_validator: don't validate packages on non-Linux
```
[this PR](https://github.com/kubernetes/system-validators/pull/12/files) added a `golang.org/x/sys/unix` import, but `unix.CGROUP2_SUPER_MAGIC` is not defined on `windows`.
cgroup and package validation only makes sense on Linux, thus split the files and no-op the cgroup and package validation for `!linux`. package validation could be extended one day to support other OSes and package managers, but might be a better idea to remove this code instead..

---

`ValidateSpec: don't include cgroups and package validators on non-Linux`

running cgroup and package validation does not make sense for non-Linux OSes.

---

`hack: run govet for all supported OS-es (darwin,linux,windows) `

This will catch compile time problems not only for the host OS
where the verification is being run.

---

### DIFF
--- a/hack/verify-govet.sh
+++ b/hack/verify-govet.sh
@@ -28,7 +28,9 @@ export GO111MODULE=on
 ERROR=0
 DIRS=$(git ls-files | grep -v "vendor\/" | grep ".go" | xargs dirname | grep -v "\." | cut -d '/' -f 1 | uniq)
 while read -r dir; do
-    go vet ./"$dir"/... || ERROR=1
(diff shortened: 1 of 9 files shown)


odinuge:
Thanks for fixing!

/lgtm

PR #15, opened 2020-04-08
revert the introduction of CgroupSpec

neolit123:
This change broke the backwards compatibility with the old SysSpec
"API". To workaround the issue without introducing a v2.0.0 release
use separate optional []string fields - e.g. CgroupsOptional,
CgroupsV2Optional.

/kind bug
/priority important-soon
/assign @liggitt


liggitt:
thanks
/lgtm

neolit123:
> do the other OS-specific files need updating?

nope, i don't think so. 
[here](https://github.com/kubernetes/system-validators/pull/14) we scoped cgroups to GOOS=linux only.

also `./hack/verify-govet.sh` should be able to catch such problems now.

liggitt:
ok, thanks

/hold cancel

PR #18, opened 2020-08-27
PidLimits promoted to GA

SergeyKanzhelev:
Follow up from https://github.com/kubernetes/kubernetes/pull/94140

### DIFF
--- a/validators/types_unix.go
+++ b/validators/types_unix.go
@@ -56,17 +56,14 @@ var DefaultSysSpec = SysSpec{
 		},
 		Forbidden: []KernelConfig{},
 	},
-	Cgroups: []string{"cpu", "cpuacct", "cpuset", "devices", "freezer", "memory"},
+	Cgroups: []string{"cpu", "cpuacct", "cpuset", "devices", "freezer", "memory", "pids"},
 	CgroupsOptional: []string{
 		// The hugetlb cgroup is optional since some kernels are compiled without support for huge pages
 		// and therefore lacks corresponding hugetlb cgroup
 		"hugetlb",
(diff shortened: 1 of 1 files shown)


neolit123:
Thanks! Lets hold this until the feature PR merges.

This also means we have to release a new Minor of this library and revendor it in k/k.

/lgtm
/approve
/hold

SergeyKanzhelev:
PR in k/k got merged. I wonder if I can cancel hold on self-authored PR...

/hold cancel

neolit123:
@SergeyKanzhelev do you happen to know if there are other planned features from SIG node related to this repository?

SergeyKanzhelev:
> @SergeyKanzhelev do you happen to know if there are other planned features from SIG node related to this repository?

I don't think so after briefly looking thru the code. even for this change, cutting release is not critical as it is not a breaking change. So perhaps, this can wait in case something else will come up

neolit123:
ok, since this library's "release process" is a bit bottlenecked on my availability, i'd prefer if we cut the release now and re-vendor in k/k right away to unblock tests if needed(?).

we cut more releases here if needed for 1.20.

neolit123:
here is the new release:
https://github.com/kubernetes/system-validators/releases/tag/v1.2.0

would you be able to send the PR to re-vendor?

here is an example commit on how we update the vendor:
https://github.com/kubernetes/kubernetes/commit/8183787493472b2d9a7006f78da63cd589456f81

it uses the script:
https://github.com/kubernetes/kubernetes/blob/master/hack/pin-dependency.sh

SergeyKanzhelev:
> here is the new release:
> https://github.com/kubernetes/system-validators/releases/tag/v1.2.0
> 
> would you be able to send the PR to re-vendor?

Yes, thank you for the detailed instructions!

PR #23, opened 2021-02-11
docker_validator: allow Docker 20.10.*

neolit123:
For Windows keep the latest version to 19.03.* as this is what
"Install-WindowsFeature -Name containers" installs on latest
versions of Windows Server (e.g. 1909 from a 2021 build on GCE).

For Linux, projects like kops have been validating 20.10 for a while.

will update this in k/k for v1.21.


odinuge (approved):
Sounds reasonable! 👍

/lgtm

PR #24, opened 2021-03-22
hack/verify-spelling: add "go mod tidy"

neolit123:
"go mod tidy" is required by go 1.16+ on new modules
to trigger the update to go.mod/sum.

without this, the verify check will fail on new PRs.


odinuge (approved):
/lgtm

PR #25, opened 2021-04-12
Add preflight check for seccomp

KentaTada:
seccomp has already been GA since Kubernetes 1.19.
I think it is essential for the current Kubernetes, but could you give me your opinion if you have any concern about adding the seccomp check?

Signed-off-by: Kenta Tada <Kenta.Tada@sony.com>

### DIFF
--- a/validators/types_unix.go
+++ b/validators/types_unix.go
@@ -57,6 +57,8 @@ var DefaultSysSpec = SysSpec{
 			{Name: "BLK_DEV_DM", Description: "Required for devicemapper."},
 			{Name: "CFS_BANDWIDTH", Description: "Required for CPU quota."},
 			{Name: "CGROUP_HUGETLB", Description: "Required for hugetlb cgroup."},
+			{Name: "SECCOMP", Description: "Required for seccomp."},
+			{Name: "SECCOMP_FILTER", Description: "Required for seccomp mode 2."},
 		},
 		Forbidden: []KernelConfig{},
 	},


validators/types_unix.go
```diff
@@ -50,6 +50,8 @@ var DefaultSysSpec = SysSpec{
 			{Name: "NETFILTER_XT_TARGET_REDIRECT", Aliases: []string{"IP_NF_TARGET_REDIRECT"}},
 			{Name: "NETFILTER_XT_MATCH_COMMENT"},
 			{Name: "FAIR_GROUP_SCHED"},
+			{Name: "SECCOMP"},
+			{Name: "SECCOMP_FILTER"},
```
neolit123, line 54: hi, given we are adding this to "required", do we know if there are any popular distributions used for k8s that have this config explicitly disabled?

KentaTada: I don't have the list of kernel configs of popular distributions although I have never heard to change the kernel config of seccomp from `y` to `n`.
I think some embedded systems may change the default to disable seccomp to reduce the impact of performance.

At first, I thought seccomp was one of requirements in the current Kubernetes.
But SeccompProfile is the optional currently and Kubernetes can launch the container without seccomp at least.
As of now, I think it is ok to change this check from `Required` to `Optional`.

KentaTada: In addition, the seccomp e2e test is not included in the conformance check.
I'm sorry I should check it at first.

KentaTada: I updated this check from `Required` to `Optional`.

neolit123:
Yes, maybe having this as optional is better given the conformance factor.
My assumption is similar - some distros may have opted out but not the
popular mainline ones.

/lgtm
/approve

PR #27, opened 2021-05-13
Split stderr and stdout in docker info parse

ironyman:
Stderr has warnings that breaks JSON unmarshal. Fix this by only reading stdout.

This is the warning
```
HOME= sudo -E docker info --format "{{json .}}"  > /dev/null
WARNING: Error loading config file: .dockercfg: $HOME is not defined
```

Looks like docker changed the way it gets HOME recently possibly this commit that's causing the warning to be emitted. 
https://github.com/docker/cli/commit/c85a37dbb472c39bc88cf707fce83fed7a644fed

In some environments, like cloud-init I think, HOME is not set. So running kubeadm in cloud-init will fail.
This fixes kubeadm init in some environments like cloud-init.

### DIFF
--- a/validators/docker_validator.go
+++ b/validators/docker_validator.go
@@ -17,6 +17,7 @@ limitations under the License.
 package system
 
 import (
+	"bytes"
 	"encoding/json"
 	"os/exec"
 	"regexp"
@@ -63,16 +64,27 @@ func (d *DockerValidator) Validate(spec SysSpec) ([]error, []error) {
 
 	// Run 'docker info' with a JSON output and unmarshal it into a dockerInfo object
 	info := dockerInfo{}
-	out, err := exec.Command("docker", "info", "--format", "{{json .}}").CombinedOutput()
+	cmd := exec.Command("docker", "info", "--format", "{{json .}}")
+
+	// Stderr can contain warnings despite docker info success.
+	var outb, errb bytes.Buffer
+	cmd.Stdout = &outb
+	cmd.Stderr = &errb
+	err := cmd.Run()
 	if err != nil {
-		return nil, []error{errors.Errorf(`failed executing "docker info --format '{{json .}}'"\noutput: %s\nerror: %v`, string(out), err)}
+		return nil, []error{errors.Errorf(`failed executing "docker info --format '{{json .}}'"\noutput: %s\nstderr: %s\nerror: %v`, outb.String(), errb.String(), err)}
 	}
-	if err := d.unmarshalDockerInfo(out, &info); err != nil {
+	if err := d.unmarshalDockerInfo(outb.Bytes(), &info); err != nil {
 		return nil, []error{err}
 	}
 
 	// validate the resulted docker info object against the spec
-	return d.validateDockerInfo(spec.RuntimeSpec.DockerSpec, info)
+	warnings, errs := d.validateDockerInfo(spec.RuntimeSpec.DockerSpec, info)
+
+	if len(errb.String()) > 0 {
+		warnings = append(warnings, errors.Errorf(`the command "docker info --format '{{json.}}'" succeeded with potential warnings\noutput: %s`, errb.String()))
+	}
+	return warnings, errs
 }
 
 func (d *DockerValidator) unmarshalDockerInfo(b []byte, info *dockerInfo) error {


KentaTada:
Hello @ironyman 
Could you sign the CLA at first?

At first glance,  you changed not to use CombinedOutput() because its function includes both stdout and stderr.
I'll review this commit later.
Related issue: https://github.com/moby/moby/issues/41890

validators/docker_validator.go
```diff
@@ -63,16 +64,27 @@ func (d *DockerValidator) Validate(spec SysSpec) ([]error, []error) {
...
+	cmd := exec.Command("docker", "info", "--format", "{{json .}}")
+
+	// Stderr can contain warnings despite docker info success.
+	var outb, errb bytes.Buffer
+	cmd.Stdout = &outb
+	cmd.Stderr = &errb
+	err := cmd.Run()
 	if err != nil {
-		return nil, []error{errors.Errorf(`failed executing "docker info --format '{{json .}}'"\noutput: %s\nerror: %v`, string(out), err)}
+			return nil, []error{errors.Errorf(`failed executing "docker info --format '{{json .}}'"\noutput: %s\nstderr: %s\nerror: %v`, outb.String(), errb.String(), err)}
```
neolit123, line 75: there is an extra tab here before "return"

validators/docker_validator.go
```diff
@@ -63,16 +64,27 @@ func (d *DockerValidator) Validate(spec SysSpec) ([]error, []error) {
...
 	}
-	if err := d.unmarshalDockerInfo(out, &info); err != nil {
-		return nil, []error{err}
+	if err := d.unmarshalDockerInfo(outb.Bytes(), &info); err != nil {
+			return nil, []error{err}
```
neolit123, line 78: there is an extra tab here before "return"

validators/docker_validator.go
```diff
@@ -63,16 +64,27 @@ func (d *DockerValidator) Validate(spec SysSpec) ([]error, []error) {
...
 	}
 
 	// validate the resulted docker info object against the spec
-	return d.validateDockerInfo(spec.RuntimeSpec.DockerSpec, info)
+	warnings, errs := d.validateDockerInfo(spec.RuntimeSpec.DockerSpec, info)
+
+	if errb.String() != "" {
+			warnings = append(warnings, errors.Errorf(`command succeeded with warning "docker info --format '{{json.}}'"\noutput: %v`, errb.String()))
```
neolit123, line 85: [1]

neolit123:
please keep the commits squashed to 1 and sign the CLA.
without the CLA we cannot accept the PR.

i think the change is good and CombinedOutput should not be used.
in terms of this actually causing failures, you should just set $HOME to mitigate.

given it's not a critical bug fix, we won't be able to backport this older kubeadm versions.

validators/docker_validator.go
ironyman: Thank you fix that.

validators/docker_validator.go
ironyman: Thank you fix that.

ironyman:
Sounds good I will do the CLA soon.

validators/docker_validator.go
```diff
@@ -63,16 +64,27 @@ func (d *DockerValidator) Validate(spec SysSpec) ([]error, []error) {
...
-	if err := d.unmarshalDockerInfo(out, &info); err != nil {
+	if err := d.unmarshalDockerInfo(outb.Bytes(), &info); err != nil {
 		return nil, []error{err}
 	}
 
 	// validate the resulted docker info object against the spec
-	return d.validateDockerInfo(spec.RuntimeSpec.DockerSpec, info)
+	warnings, errs := d.validateDockerInfo(spec.RuntimeSpec.DockerSpec, info)
+
+	if errb.String() != "" {
```
neolit123, line 84:
```suggestion
	if len(errb.String()) > 0 {
```
this is just a styling change.

validators/docker_validator.go
```diff
@@ -63,16 +64,27 @@ func (d *DockerValidator) Validate(spec SysSpec) ([]error, []error) {
...
+		warnings = append(warnings, errors.Errorf(`the command "docker info --format '{{json.}}'" succeeded with warning\noutput: %v`, errb.String()))
```
neolit123, line 85:
```suggestion
		warnings = append(warnings, errors.Errorf(`the command "docker info --format '{{json.}}'" succeeded with warning\noutput: %s`, errb.String()))
```

neolit123 (commented):
/ok-to-test
/approve

the change looks mostly fine, added a couple of nits.

neolit123:
please keep the commits squashed to 1.

1.22 will be in code freeze so if you wish this change to be part of it we must tag a new PATCH version in this repo and re-vendor it in k8s. would you have the time to do that?
https://github.com/kubernetes/system-validators/tree/ca2e257c13e6d6e49e0de2a31d08c03c790f548f#creating-releases

see the `pin-dependency.sh` part.

neolit123:
actually we need to release a new version due to https://github.com/kubernetes/system-validators/commit/c9ffc95d14808b10c7d2780eedd3cfcf94dca869

@KentaTada is this correct?

KentaTada:
> actually we need to release a new version due to [c9ffc95](https://github.com/kubernetes/system-validators/commit/c9ffc95d14808b10c7d2780eedd3cfcf94dca869)
> 
> @KentaTada is this correct?

Yes. I think the preflight check for seccomp is needed for the current k8s.

ironyman:
So we should create a new tag v.1.4.1 on master? What do we have to do for preflight check?

neolit123:
done v1.5.0 is now created. we can now update kubernetes/kubernetes to use it for 1.22.
https://github.com/kubernetes/system-validators/releases/tag/v1.5.0

ironyman:
Should I make a PR in kubernetes repo after the pin-dependency.sh? Anything I should do to validate the changes in kubernetes repo?

neolit123:
yes, if you can send the PR that would be great.

here is how the 1.4.0 PR looked like https://github.com/kubernetes/kubernetes/pull/98977
you add a similar title / description.

to verify the change you can:
- pin the dependency to v1.5.0
- use `make all WHAT=cmd/kubeadm` to build kubeadm
- find the kubeadm binary in the /output folder
- check if the kubeadm binary exhibits the same problem that you tried to fix in this current PR

neolit123:
> 1.22 will be in code freeze soon (early July)

the date is July 8th (next week Thursday). at that point this 1.5.0 update will not be accepted for 1.22 and has to wait for 1.23.

PR #29, opened 2022-02-22
Add blkio for iops limit support in the future

### LINKED ISSUE #28: kernel cgroup v2 support update
https://github.com/kubernetes/system-validators/blob/73d9d3d9f43503b1a64ec0506080252c683f8408/validators/types_unix.go#L65-L77

Need we update it?

- blkio: As I know, iops cgroupv2 needs `blkio` once https://github.com/containerd/containerd/pull/5490 is ready. 

like hugetlb: huge page support is optional.

pacoxu:
Fixes #28 

The support of iops in kubernetes and contained can be found and tracked in https://github.com/kubernetes/kubernetes/issues/92287#issuecomment-1010723587 and https://github.com/kubernetes/enhancements/pull/1907, https://github.com/containerd/containerd/pull/5490, https://github.com/cri-o/cri-o/pull/4873.

### DIFF
--- a/validators/cgroup_validator_linux.go
+++ b/validators/cgroup_validator_linux.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 /*
--- a/validators/types_unix.go
+++ b/validators/types_unix.go
@@ -1,3 +1,4 @@
+//go:build !windows
 // +build !windows
 
 /*
--- a/validators/types_unix.go
+++ b/validators/types_unix.go
@@ -67,9 +67,12 @@ var DefaultSysSpec = SysSpec{
 		// The hugetlb cgroup is optional since some kernels are compiled without support for huge pages
 		// and therefore lacks corresponding hugetlb cgroup
 		"hugetlb",
+		// The blkio cgroup is optional since some kernels are compiled without support for block I/O throttling.
+		// Containerd and cri-o will use blkio to track disk I/O and throttling in both cgroup v1 and v2.
+		"blkio",
 	},
 	CgroupsV2:         []string{"cpu", "cpuset", "devices", "freezer", "memory", "pids"},
-	CgroupsV2Optional: []string{"hugetlb"},
+	CgroupsV2Optional: []string{"hugetlb", "blkio"},
 	RuntimeSpec: RuntimeSpec{
 		DockerSpec: &DockerSpec{
 			Version:     []string{`1\.1[1-3]\..*`, `17\.0[3,6,9]\..*`, `18\.0[6,9]\..*`, `19\.03\..*`, `20\.10\..*`},
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module k8s.io/system-validators
 
-go 1.13
+go 1.16
 
 require (
 	github.com/blang/semver v3.5.0+incompatible
--- a/validators/cgroup_validator_other.go
+++ b/validators/cgroup_validator_other.go
@@ -1,3 +1,4 @@
+//go:build !linux
 // +build !linux
 
 /*
--- a/validators/cgroup_validator_test.go
+++ b/validators/cgroup_validator_test.go
@@ -1,3 +1,4 @@
+//go:build linux
 // +build linux
 
 /*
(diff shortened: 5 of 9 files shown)


validators/cgroup_validator_linux.go
```diff
@@ -1,3 +1,4 @@
+//go:build linux
```
neolit123, line 1: with this change are we bumping the minimum required go version?
https://github.com/kubernetes/system-validators/blob/master/go.mod#L3

validators/types_unix.go
```diff
@@ -66,10 +67,10 @@ var DefaultSysSpec = SysSpec{
 	CgroupsOptional: []string{
 		// The hugetlb cgroup is optional since some kernels are compiled without support for huge pages
 		// and therefore lacks corresponding hugetlb cgroup
-		"hugetlb",
+		"hugetlb", "blkio",
```
neolit123, line 70: we should have this on a separate line and have a description comment similarly to the one for hugetlb.

validators/types_unix.go
```diff
@@ -66,10 +67,10 @@ var DefaultSysSpec = SysSpec{
...
 	},
 	CgroupsV2:         []string{"cpu", "cpuset", "devices", "freezer", "memory", "pids"},
-	CgroupsV2Optional: []string{"hugetlb"},
+	CgroupsV2Optional: []string{"hugetlb", "blkio"},
```
neolit123, line 73: what about cgroups v1? are the container runtimes going to support this with v1 as well?
if not, this can be explained in the comment as well:
https://github.com/kubernetes/system-validators/pull/29/files#r812016832

neolit123 (commented):
added a few comments @pacoxu 
i don't object to the change but we need to be clear about the scope.

before the 1.24 release we can create a new release of this library and vendor it in k/k.

validators/cgroup_validator_linux.go
pacoxu: bump the go version to 1.16?

validators/types_unix.go
pacoxu: Add some description now.

validators/cgroup_validator_linux.go
neolit123: Can we avoid changing these go directives? Did you get an error without the change?

validators/types_unix.go
```diff
@@ -67,9 +68,12 @@ var DefaultSysSpec = SysSpec{
 		// The hugetlb cgroup is optional since some kernels are compiled without support for huge pages
 		// and therefore lacks corresponding hugetlb cgroup
 		"hugetlb",
+		// The blkio cgroup is optional since some kernels are compiled without support for block I/O throttling.
+		// Containerd and cri-o will use blkio to track disk I/O and throttling in both cgroup v1 and v2.
+		"blkio",
```
neolit123, line 73: If the CRs are going to also track v1, we should add it under v1 as optional here ad well?

pacoxu: Yes, I add it to both v1 and v2.

- CgroupsOptional is for v1
- CgroupsV2Optional is for v2

like hugetlb, I add comments to explain in v1 only.

validators/cgroup_validator_linux.go
neolit123: @pacoxu this question? ^

validators/types_unix.go
neolit123: ok, SGTM.

neolit123 (commented):
the code changes LGTM, minus this comment about minimum go version:
https://github.com/kubernetes/system-validators/pull/29#discussion_r814025032

if we can't avoid bumping to 1.16 it's fine.

pacoxu:
If I drop the go version bumping(with golang format) commit, the verify ci will fail like https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/system-validators/29/pull-system-validators-verify/1497038772368314368.

pacoxu:
```
root@pacoall ~]# docker run -it --entrypoint=bash  gcr.io/k8s-staging-test-infra/kubekins-e2e:v20220221-c13e827224-master
root@427ff46eeefa:/workspace# go version
go version go1.17.7 linux/amd64
```

neolit123 (commented):
Ok, understood. This is fine.
/lgtm
/approve

PR #30, opened 2022-03-11
Upgrade to blang/semver/v4 v4.0.0

skitt:
This is the first version to be fully go-mod-compatible.

Signed-off-by: Stephen Kitt <skitt@redhat.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module k8s.io/system-validators
 go 1.16
 
 require (
-	github.com/blang/semver v3.5.0+incompatible
+	github.com/blang/semver/v4 v4.0.0
 	github.com/pkg/errors v0.8.0
 	github.com/stretchr/testify v1.3.0
 	golang.org/x/sys v0.0.0-20191022100944-742c48ecaeb7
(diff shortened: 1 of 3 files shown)


neolit123 (commented):
Looks like it is only used in the package validator. I think kubeadm is not running the package checks at least.

Are you using this library and saw a problem due to the semver library not being at v4?

skitt:
> Looks like it is only used in the package validator. I think kubeadm is not running the package checks at least.
> 
> Are you using this library and saw a problem due to the semver library not being at v4?

Not directly; this is for consistency with https://github.com/kubernetes/kubernetes/pull/108629, to clean up the dependency tree of dependencies further down from Kubernetes (where we end up with both version 3.5.1 and 4.0.0 of `semver`).

neolit123:
/ok-to-test

Ok that is fair, given k/k is the main user but this lib will have to be bumped with a new release and revendored in k/k.

There is already one MINOR level feature pending release. Releases of this lib happen close before code freeze of a k/k release (pending).

If there was nothing else commited we could have made a patch with the semver bump.

neolit123:
/lgtm
/approve

i should keep a note to release a new version of this repository before the 30th of march (code freeze for k8s 1.24).

skitt:
> i should keep a note to release a new version of this repository before the 30th of march (code freeze for k8s 1.24).

Gentle reminder about this so there’s time to merge the bump in k/k itself...

neolit123:
had a todo for next week, but i guess i can push the release today.
don't think anything new will land here before the k8s CF.

skitt:
Thanks!

Issue #31, opened 2022-05-30
kubeadm shows SystemVerification warning about "missing optional cgroups: blkio"


orimanabu:
**Sysmptom**
I ran kubeadm on CentOS Stream 9 and got a warning.

```
	[WARNING SystemVerification]: missing optional cgroups: blkio
```

**Reproducibility**
Always

**Environment**
- OS: CentOS Stream 9
- Kubernetes: v1.24.1
- CRI-O: v1.24.0

**Investigation**
CentOS Stream 9 runs with cgroups v2 only.
"blkio" controller seems to be available only for cgroups v1 according to [man cgroups(7)](https://man7.org/linux/man-pages/man7/cgroups.7.html).

```
   Cgroups v2 controllers
<snip>
       io (since Linux 4.5)
              This is the successor of the version 1 blkio controller.
```

We should update CgroupsV2Optional for cgroups v2 ?

neolit123:
IIRC This was intended like so. It applies to v1 and v2. 
Cc @pacoxu

orimanabu:
FYI, this is my `/sys/fs/cgroup/cgroup.controllers`.
```
[ori@c9s ~]$ cat /sys/fs/cgroup/cgroup.controllers 
cpuset cpu io memory hugetlb pids rdma misc
```

```
[ori@c9s ~]$ cat /etc/os-release 
NAME="CentOS Stream"
VERSION="9"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="9"
PLATFORM_ID="platform:el9"
PRETTY_NAME="CentOS Stream 9"
ANSI_COLOR="0;31"
LOGO="fedora-logo-icon"
CPE_NAME="cpe:/o:centos:centos:9"
HOME_URL="https://centos.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux 9"
REDHAT_SUPPORT_PRODUCT_VERSION="CentOS Stream"
[ori@c9s ~]$ uname -r
5.14.0-96.el9.x86_64
```

KentaTada:
Thanks! @orimanabu 
I took a quick look at this issue.
When system-validators gets the list of controllers are enabled by the kernel, it reads `the cgroup mount point + "cgroup.controllers"`(Ex. /sys/fs/cgroup/cgroup.controllers).
[...]

PR #32, opened 2022-05-30
Use not "blkio" but "io" controller in cgroups v2

orimanabu:
"blkio" controller seems to be available only for cgroups v1 according to [man cgroups(7)](https://man7.org/linux/man-pages/man7/cgroups.7.html).

See also: https://github.com/kubernetes/system-validators/issues/31

### DIFF
--- a/validators/types_unix.go
+++ b/validators/types_unix.go
@@ -72,8 +72,12 @@ var DefaultSysSpec = SysSpec{
 		// Containerd and cri-o will use blkio to track disk I/O and throttling in both cgroup v1 and v2.
 		"blkio",
 	},
-	CgroupsV2:         []string{"cpu", "cpuset", "devices", "freezer", "memory", "pids"},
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
/approve
/cc @KentaTada @pacoxu 

PTAL for LGTM
if we merge this it can be considered a bug fix, correct?

/kind bug

We can include it as a PATCH semver here, but backporting the vendor to k/k 1.24 doesn't seem necessary given users can skip the warning. Thoughts?

pacoxu (approved):
/lgtm
https://www.kernel.org/doc/Documentation/cgroup-v2.txt
```
The "io" controller regulates the distribution of IO resources. 
```
It's my fault. Thanks for the fix.

neolit123:
looks like there is a gofmt error in CI.

orimanabu:
Thank you for merging this PR!
This is my first contribution to Kubernetes.

PR #35, opened 2023-03-21
[WIP] Warn kernel 4.0 

pacoxu:
xref https://github.com/kubernetes/kubernetes/issues/116799 as there are still no conclusions.

### DIFF
--- a/validators/kernel_validator_test.go
+++ b/validators/kernel_validator_test.go
@@ -32,54 +32,68 @@ func TestValidateKernelVersion(t *testing.T) {
 	// This is fine, because the test mainly tests the kernel version validation logic,
 	// not the DefaultSysSpec. The DefaultSysSpec should be tested with node e2e.
 	testRegex := []string{`^3\.[1-9][0-9].*$`, `^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}
+	testSuggetRegex := []string{`^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}
 	for _, test := range []struct {
 		name    string
 		version string
 		err     bool
+		warn    bool
 	}{
 		{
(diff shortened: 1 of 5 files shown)


validators/kernel_validator_test.go
```diff
@@ -32,55 +32,69 @@ func TestValidateKernelVersion(t *testing.T) {
...
 			if !test.err {
 				assert.Nil(t, err, "Expect error not to occur with kernel version %q", test.version)
 			} else {
 				assert.NotNil(t, err, "Expect error to occur with kenrel version %q", test.version)
 			}
+			if !test.warn {
+				assert.Nil(t, warn, "Expect warn not to occur with kernel version %q", test.version)
+			} else {
+				assert.NotNil(t, warn, "Expect warn to occur with kenrel version %q", test.version)
+			}
```
neolit123, line 97:
```suggestion
			if !test.warn {
				assert.Nil(t, warn, "Expect warning not to occur with kernel version %q", test.version)
			} else {
				assert.NotNil(t, warn, "Expect warning to occur with kernel version %q", test.version)
			}
```

neolit123: there is "kenrel" typo in the old code too.

neolit123 (commented):
the Versions / SuggestedVersions separation seems OK.
ideally should not have it and just say "version >= X" is required.
it might be a bit confusing for users what it means to have a "suggested version". they might not know what does not work.

either way, let's wait for comments by sig node on the linked issue.

pacoxu:
> it might be a bit confusing for users what it means to have a "suggested version". they might not know what does not work.

I renamed it from suggested to recommended. And I add RecommendedNote to make the message more readable.

pacoxu:
/close

I may work on this once cgroup v1 is not supported and iptables is not supported oneday. ❓

PR #37, opened 2024-07-24
kernel version should be 4.19+; recommend version 5.8 for cgroup v2

pacoxu:
For kernel long term support, see https://wiki.linuxfoundation.org/civilinfrastructureplatform/start and https://endoflife.date/linux
-  4.4 & 4.19 are selected as kernel Super Long Term Support (SLTS), and the Civil Infrastructure Platform will provide support until at least 2026. 
   -  **EDITED removed 4.4** as CIP is not in https://www.kernel.org/category/releases.html and maintained in a different place.
- For [cgroup v2](https://kubernetes.io/docs/concepts/architecture/cgroups/), Kubernetes recommends to use 5.8 and later, and in runc docs, the minimal version is 4.15 and 5.2+ is recommended. 

Other comments that may be related:

> In Kubernetes 1.31, cgroup v1 is moved to maintenance mode and 4.14 LTS EOF in Jan 2024 (linux, LTS) , besides, centos 7 is EOL in June 30, 2024. I chosen 4.15 as runc.
> 
> - for kernel minimal version, choosing 4.15 as runc + cgroup v2 https://github.com/opencontainers/runc/blob/3778ae603c706494fd1e2c2faf83b406e38d687d/docs/cgroup-v2.md?plain=1#L24
>   - kernel >= 4.15 with CONFIG_CGROUP_DEVICE and CONFIG_CGROUP_BPF is required.  From https://github.com/containerd/containerd/pull/3799#issuecomment-555740694. 
>   - cpu (since Linux 4.15)
> - cgroup v2: Kubernetes recommended kernel version https://kubernetes.io/docs/concepts/architecture/cgroups/
> 
> More details can be found in https://github.com/kubernetes/kubernetes/issues/116799.
> 
> The v1.31 KEP https://github.com/kubernetes/enhancements/issues/4569
> -  https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/4569-cgroup-v1-maintenance-mode/README.md KEP-4569: Moving cgroup v1 support into maintenance mode 
> 
> Other minimal kernel version candidates
> - kernel 4.5 announced that cgroup v2 is not experimental anymore, as it supports io/pids/memory.
> - runc recommends 5.2+ as 5.2 supports freezer.

### DIFF
--- a/validators/kernel_validator.go
+++ b/validators/kernel_validator.go
@@ -90,7 +90,7 @@ func (k *KernelValidator) validateKernelVersion(kSpec KernelSpec) error {
 		}
 	}
 	k.Reporter.Report("KERNEL_VERSION", k.kernelRelease, bad)
-	return fmt.Errorf("unsupported kernel release: %s", k.kernelRelease)
+	return fmt.Errorf("kernel release %s is unsupported. %s", k.kernelRelease, kSpec.VersionsNote)
 }
 
 // validateKernelConfig validates the kernel configurations.
--- a/validators/kernel_validator_test.go
+++ b/validators/kernel_validator_test.go
@@ -31,20 +31,45 @@ func TestValidateKernelVersion(t *testing.T) {
 	// they may be different.
 	// This is fine, because the test mainly tests the kernel version validation logic,
 	// not the DefaultSysSpec. The DefaultSysSpec should be tested with node e2e.
-	testRegex := []string{`^3\.[1-9][0-9].*$`, `^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}
+	testRegex := []string{`^4\.19.*$`, `^4\.[2-9][0-9].*$`, `^([5-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}
 	for _, test := range []struct {
 		name    string
 		version string
 		err     bool
 	}{
 		{
-			name:    "3.19.9-99-test first version regex matches",
+			name:    "3.19.9-99-test no version regex matches",
 			version: "3.19.9-99-test",
+			err:     true,
+		},
+		{
+			name:    "4.4.14+ no version regex matches",
+			version: "4.4.14+",
+			err:     true,
+		},
+		{
+			name:    "4.7.1 no version regex matches",
+			version: "4.7.1",
+			err:     true,
+		},
+		{
+			name:    "4.17.3 no version regex matches",
+			version: "4.17.3",
+			err:     true,
+		},
+		{
+			name:    "4.19.3-99-test matches",
+			version: "4.19.3-99-test",
 			err:     false,
 		},
 		{
-			name:    "4.4.14+ one of version regexes matches",
-			version: "4.4.14+",
+			name:    "4.20.3+ matches",
+			version: "4.20.3+",
+			err:     false,
+		},
+		{
+			name:    "5.12.3 matches",
+			version: "5.12.3",
 			err:     false,
 		},
 		{
@@ -79,7 +104,7 @@ func TestValidateKernelVersion(t *testing.T) {
 			if !test.err {
 				assert.Nil(t, err, "Expect error not to occur with kernel version %q", test.version)
 			} else {
-				assert.NotNil(t, err, "Expect error to occur with kenrel version %q", test.version)
+				assert.NotNil(t, err, "Expect error to occur with kernel version %q", test.version)
 			}
 		})
 	}
@@ -189,7 +214,7 @@ func TestValidateCachedKernelConfig(t *testing.T) {
 			if !test.err {
 				assert.Nil(t, err, "Expect error not to occur with kernel config %q", test.config)
 			} else {
-				assert.NotNil(t, err, "Expect error to occur with kenrel config %q", test.config)
+				assert.NotNil(t, err, "Expect error to occur with kernel config %q", test.config)
 			}
 		})
 	}
--- a/validators/types_unix.go
+++ b/validators/types_unix.go
@@ -28,7 +28,9 @@ import (
 var DefaultSysSpec = SysSpec{
 	OS: "Linux",
 	KernelSpec: KernelSpec{
-		Versions: []string{`^3\.[1-9][0-9].*$`, `^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 3.10+, or newer
+		// 4.19 is an active kernel Long Term Support (LTS) release, tracked in https://www.kernel.org/category/releases.html.
+		Versions:     []string{`^4\.19.*$`, `^4\.[2-9][0-9].*$`, `^([5-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`},
+		VersionsNote: "Recommended LTS version from the 4.x series is 4.19. Any 5.x or 6.x versions are also supported. For cgroups v2 support, the minimal version is 4.15 and the recommended version is 5.8+",
 		// TODO(random-liu): Add more config
 		// TODO(random-liu): Add description for each kernel configuration:
 		Required: []KernelConfig{
--- a/validators/types_windows.go
+++ b/validators/types_windows.go
@@ -28,10 +28,11 @@ import (
 var DefaultSysSpec = SysSpec{
 	OS: "Microsoft Windows Server 2016",
 	KernelSpec: KernelSpec{
-		Versions:  []string{`10\.0\.1439[3-9]`, `10\.0\.14[4-9][0-9]{2}`, `10\.0\.1[5-9][0-9]{3}`, `10\.0\.[2-9][0-9]{4}`, `10\.[1-9]+\.[0-9]+`}, //requires >= '10.0.14393'
-		Required:  []KernelConfig{},
-		Optional:  []KernelConfig{},
-		Forbidden: []KernelConfig{},
+		Versions:     []string{`10\.0\.1439[3-9]`, `10\.0\.14[4-9][0-9]{2}`, `10\.0\.1[5-9][0-9]{3}`, `10\.0\.[2-9][0-9]{4}`, `10\.[1-9]+\.[0-9]+`}, //requires >= '10.0.14393'
+		VersionsNote: "The kernel version should be >= '10.0.14393'",
+		Required:     []KernelConfig{},
+		Optional:     []KernelConfig{},
+		Forbidden:    []KernelConfig{},
 	},
 	RuntimeSpec: RuntimeSpec{
 		DockerSpec: &DockerSpec{
--- a/validators/types.go
+++ b/validators/types.go
@@ -42,6 +42,8 @@ type KernelConfig struct {
 type KernelSpec struct {
 	// Versions define supported kernel version. It is a group of regexps.
 	Versions []string `json:"versions,omitempty"`
+	// VersionsNote provides additional information if Versions do not match.
+	VersionsNote string `json:"versionsNote,omitempty"`
 	// Required contains all kernel configurations required to be enabled
 	// (built in or as module).
 	Required []KernelConfig `json:"required,omitempty"`


validators/kernel_validator.go
```diff
@@ -67,30 +67,52 @@ func (k *KernelValidator) Validate(spec SysSpec) ([]error, []error) {
 	}
 	k.kernelRelease = release
 	var errs []error
-	if err = k.validateKernelVersion(spec.KernelSpec); err != nil {
+	var warns []error
```
neolit123, line 70: should be:
```
var (
...
)
```

validators/kernel_validator.go
```diff
@@ -67,30 +67,52 @@ func (k *KernelValidator) Validate(spec SysSpec) ([]error, []error) {
...
+	return warns, errs
 }
 
 // validateKernelVersion validates the kernel version.
-func (k *KernelValidator) validateKernelVersion(kSpec KernelSpec) error {
-	versionRegexps := kSpec.Versions
-	for _, versionRegexp := range versionRegexps {
+func (k *KernelValidator) validateKernelVersion(kSpec KernelSpec) (error, error) {
+	var warn, err error
+	var matched bool
```
neolit123, line 90: same here. better to have a `var (...) ` block.

validators/kernel_validator_test.go
```diff
@@ -32,54 +32,92 @@ func TestValidateKernelVersion(t *testing.T) {
...
 		},
 		{
-			name:    "4.4.14+ one of version regexes matches",
+			name:    "4.4.14+ no version regex matches",
 			version: "4.4.14+",
 			err:     false,
+			warn:    true,
+		},
+		{
+			name:    "5.7.3  no version regex matches",
```
neolit123, line 55:
```suggestion
			name:    "5.7.3 no version regex matches",
```

validators/types_unix.go
```diff
@@ -28,7 +28,12 @@ import (
 var DefaultSysSpec = SysSpec{
 	OS: "Linux",
 	KernelSpec: KernelSpec{
-		Versions: []string{`^3\.[1-9][0-9].*$`, `^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 3.10+, or newer
+		// runc cgroup v2 https://github.com/opencontainers/runc/blob/3778ae603c706494fd1e2c2faf83b406e38d687d/docs/cgroup-v2.md?plain=1#L24
+		Versions: []string{`^3\.[1-9][0-9].*$`, `^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 4.15+, or newer
+		// cgroup v2: Kubernetes recommended kernel version
+		// https://kubernetes.io/docs/concepts/architecture/cgroups/
+		RecommendedVersions: []string{`^5\.[8-9].*$`, `^5\.[1-9][0-9].*$`, `^([6-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 5.8+, or newer for cgroup v2
+		RecommendedNote:     "kernel version should >= '4.15', and 5.8+ is recommended.",
```
neolit123, line 36: can be called `VersionsNote`

from a discussion on slack we had:
> the checks are regexp based so could be 4.4 == ok, 4.19 == ok, any other >= 4.19 version == ok, any other version == err.

so  in this note we can say that all those 4.x LTS versions are supported but the recommendation is to use 4.15 or 5.8+.

neolit123 (commented):
we could show an error for any version that is not in that list of Versions and we don't need a RecommendedVersions or warnings IMO.
the error message could be x is not in the list of recommended versions. %s, where %s is a VersionsNote from the KernelSpec API.

validators/types_unix.go
pacoxu:
```
VersionsNote: "kernel version should be in maintenance, at least 4.4 or 4.19+, and to use cgroup v2, 5.8+ is recommended.",
```

update it as above..

validators/kernel_validator.go
```diff
@@ -90,7 +90,7 @@ func (k *KernelValidator) validateKernelVersion(kSpec KernelSpec) error {
 		}
 	}
 	k.Reporter.Report("KERNEL_VERSION", k.kernelRelease, bad)
-	return fmt.Errorf("unsupported kernel release: %s", k.kernelRelease)
+	return fmt.Errorf("kernel release %s is unsupported, and %s", k.kernelRelease, kSpec.VersionsNote)
```
pacoxu, line 93: The final error message is like`kernel release %s is unsupported, and kernel version should be in maintenance, at least 4.4 or 4.19+, and to use cgroup v2, 5.8+ is recommended.`

neolit123: i think we can make it
```suggestion
	return fmt.Errorf("kernel release %s is unsupported. %s", k.kernelRelease, kSpec.VersionsNote)
```

so just put `.` and after that is the note.

validators/types_unix.go
```diff
@@ -28,7 +28,10 @@ import (
 var DefaultSysSpec = SysSpec{
 	OS: "Linux",
 	KernelSpec: KernelSpec{
-		Versions: []string{`^3\.[1-9][0-9].*$`, `^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 3.10+, or newer
+		// 4.4 & 4.19 are selected as kernel Super Long Term Support (SLTS), and the Civil Infrastructure Platform will provide support until at least 2026.
+		// For cgroup v2, Kubernetes recommends to use 5.8 and later, and in runc docs, the minimal version is 4.15 and 5.2+ is recommended.
+		Versions:     []string{`^4\.4.*$`, `^4\.19.*$`, `^4\.[2-9][0-9].*$`, `^([5-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 5.8+, or newer for cgroup v2
+		VersionsNote: "kernel version should be in maintenance, at least 4.4 or 4.19+, and to use cgroup v2, 5.8+ is recommended.",
```
neolit123, line 34:
```suggestion
		VersionsNote: "Recommended LTS versions from the 4.x series are 4.4 and 4.19. Any 4.x version newer than 4.19 is also supported. For cgroups v2 support, the minimal version is 4.15 and the recommended version is 5.8+",
```

adding a more verbose message.

validators/types_windows.go
```diff
@@ -28,10 +28,11 @@ import (
 var DefaultSysSpec = SysSpec{
 	OS: "Microsoft Windows Server 2016",
 	KernelSpec: KernelSpec{
-		Versions:  []string{`10\.0\.1439[3-9]`, `10\.0\.14[4-9][0-9]{2}`, `10\.0\.1[5-9][0-9]{3}`, `10\.0\.[2-9][0-9]{4}`, `10\.[1-9]+\.[0-9]+`}, //requires >= '10.0.14393'
-		Required:  []KernelConfig{},
-		Optional:  []KernelConfig{},
-		Forbidden: []KernelConfig{},
+		Versions:     []string{`10\.0\.1439[3-9]`, `10\.0\.14[4-9][0-9]{2}`, `10\.0\.1[5-9][0-9]{3}`, `10\.0\.[2-9][0-9]{4}`, `10\.[1-9]+\.[0-9]+`}, //requires >= '10.0.14393'
+		VersionsNote: "kernel version should >= '10.0.14393'",
```
neolit123, line 32:
```suggestion
		VersionsNote: "The kernel version should be >= '10.0.14393'",
```

validators/types_unix.go
```diff
@@ -28,7 +28,10 @@ import (
 var DefaultSysSpec = SysSpec{
 	OS: "Linux",
 	KernelSpec: KernelSpec{
-		Versions: []string{`^3\.[1-9][0-9].*$`, `^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 3.10+, or newer
+		// 4.4 & 4.19 are selected as kernel Super Long Term Support (SLTS), and the Civil Infrastructure Platform will provide support until at least 2026.
+		// For cgroup v2, Kubernetes recommends to use 5.8 and later, and in runc docs, the minimal version is 4.15 and 5.2+ is recommended.
+		Versions:     []string{`^4\.4.*$`, `^4\.19.*$`, `^4\.[2-9][0-9].*$`, `^([5-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 5.8+, or newer for cgroup v2
```
neolit123, line 33: please move the comment or remove it:
> // Requires 5.8+, or newer for cgroup v2.

i think there is a disagreement between the comments.
is 5.2+ or 5.8+ recommended for cgroup v2?

validators/kernel_validator_test.go
```diff
@@ -31,22 +31,42 @@ func TestValidateKernelVersion(t *testing.T) {
...
-			name:    "4.4.14+ one of version regexes matches",
+			name:    "4.4.14+ no version regex matches",
 			version: "4.4.14+",
 			err:     false,
 		},
+		{
+			name:    "4.17.3 no version regex matches",
+			version: "4.17.3",
+			err:     true,
+		},
```
neolit123, line 54: can you please also add a test case for 4.6 or something else that is between the LTS versions and returns an error?

validators/types_unix.go
pacoxu: We may take kubernetes 5.8 here. 

```
// Requires 5.8+, or newer for cgroup v2.
```

validators/kernel_validator_test.go
pacoxu: Added.

validators/types_unix.go
neolit123: can we remove the extra comment on this line:
````
Versions:     []string{`^4\.4.*$`, `^4\.19.*$`, `^4\.[2-9][0-9].*$`, `^([5-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 5.8+, or newer for cgroup v2
```

neolit123 (commented):
just one final nit 
https://github.com/kubernetes/system-validators/pull/37/files#r1691979088
and LGTM

pacoxu:
Updated.

neolit123 (commented):
/lgtm
/approve

should we create a release after this PR merges?

neolit123:
/hold
drop the hold if needed.

pacoxu:
> should we create a release after this PR merges?

Yes. 


> /hold
> drop the hold if needed.

We may wait for another week or two in case SIG-Node owners have some inputs.

neolit123:
> > should we create a release after this PR merges?
> 
> Yes.
> 
> > /hold
> > drop the hold if needed.
> 
> We may wait for another week or two in case SIG-Node owners have some inputs.

you should probably give them a ping on slack, or they will miss this.

validators/types_unix.go
```diff
@@ -28,7 +28,9 @@ import (
 var DefaultSysSpec = SysSpec{
 	OS: "Linux",
 	KernelSpec: KernelSpec{
-		Versions: []string{`^3\.[1-9][0-9].*$`, `^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 3.10+, or newer
+		// 4.4 & 4.19 are selected as kernel Super Long Term Support (SLTS), and the Civil Infrastructure Platform will provide support until at least 2026.
+		Versions:     []string{`^4\.4.*$`, `^4\.19.*$`, `^4\.[2-9][0-9].*$`, `^([5-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`},
+		VersionsNote: "Recommended LTS versions from the 4.x series are 4.4 and 4.19. Any 4.x version newer than 4.19 is also supported. For cgroups v2 support, the minimal version is 4.15 and the recommended version is 5.8+",
```
SergeyKanzhelev, line 33: for this note, do we have a check on cgroupv2 that the minimal version is satisfied?

cc: @harche

pacoxu: The current statement is from 
1. https://github.com/opencontainers/runc/blob/ad5b481dace5cda8ca7c659b7717a15517333198/docs/cgroup-v2.md?plain=1#L23-L24 
2.  https://kubernetes.io/docs/concepts/architecture/cgroups/#requirements: `Linux Kernel version is 5.8 or later`

If the runc one is not accurate, we may change to `the minimal version is 5.8`

validators/kernel_validator_test.go
```diff
@@ -31,22 +31,47 @@ func TestValidateKernelVersion(t *testing.T) {
...
 		{
-			name:    "3.19.9-99-test first version regex matches",
+			name:    "3.19.9-99-test no version regex matches",
 			version: "3.19.9-99-test",
-			err:     false,
+			err:     true,
 		},
 		{
-			name:    "4.4.14+ one of version regexes matches",
+			name:    "4.4.14+ no version regex matches",
```
SataQiu, line 46:
```suggestion
			name:    "4.4.14+ first version regex matches",
```

SataQiu (approved):

validators/types_unix.go
SergeyKanzhelev: I'm fine with the message. Just was wondering if we need to add a validaiton in kubelet on start when cgroupv2 are used.

pacoxu: there are many kernel version validations in kube. Mainly related to some FG or sysctl.

We may add a warning first for kubelet using cgoup v2.

validators/kernel_validator_test.go
pacoxu: Updated.

pacoxu:
According to some discussions in https://github.com/endoflife-date/endoflife.date/pull/5608#issuecomment-2268591430, the CIP seems to be not in https://www.kernel.org/category/releases.html and maintained in a different place(gitlab) than `torvalds/linux`.  

I prefer to remove the 4.4.* here.

neolit123:
> According to some discussions in [endoflife-date/endoflife.date#5608 (comment)](https://github.com/endoflife-date/endoflife.date/pull/5608#issuecomment-2268591430), the CIP seems to be not in https://www.kernel.org/category/releases.html and maintained in a different place(gitlab) than `torvalds/linux`.
> 
> I prefer to remove the 4.4.* here.

+1 to keep only official lts

pacoxu:
Updated in commit: [update according to kernel official active kernel releases](https://github.com/kubernetes/system-validators/pull/37/commits/020ec91cd005724e9625d632dade5c4550a9fa6c).

validators/types_unix.go
```diff
@@ -28,7 +28,9 @@ import (
 var DefaultSysSpec = SysSpec{
 	OS: "Linux",
 	KernelSpec: KernelSpec{
-		Versions: []string{`^3\.[1-9][0-9].*$`, `^([4-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`}, // Requires 3.10+, or newer
+		// 4.19 is active kernel Long Term Support (LTS) release in https://www.kernel.org/category/releases.html.
```
neolit123, line 31:
```suggestion
		// 4.19 is an active kernel Long Term Support (LTS) release, tracked in https://www.kernel.org/category/releases.html.
```

validators/types_unix.go
```diff
@@ -28,7 +28,9 @@ import (
...
+		Versions:     []string{`^4\.19.*$`, `^4\.[2-9][0-9].*$`, `^([5-9]|[1-9][0-9]+)\.([0-9]+)\.([0-9]+).*$`},
+		VersionsNote: "Recommended LTS versions from the 4.x series is 4.19. Any 5.x or 6.x versions are also supported. For cgroups v2 support, the minimal version is 4.15 and the recommended version is 5.8+",
```
neolit123, line 33:
```suggestion
		VersionsNote: "Recommended LTS version from the 4.x series is 4.19. Any 5.x or 6.x versions are also supported. For cgroups v2 support, the minimal version is 4.15 and the recommended version is 5.8+",
```

neolit123 (commented):
please squash the commits to 1 as well.

neolit123 (commented):
/lgtm

validators/types_unix.go
pacoxu: https://github.com/kubernetes/kubernetes/pull/126595

pacoxu:
/unhold
@neolit123  Could you trigger a release after v1.31.0 is released this week.

PR #38, opened 2024-07-24
add cgroup v1 warning as it is in maintaince mode for kube

pacoxu:
- add a warning if the node is cgroup v1

The v1.31 KEP https://github.com/kubernetes/enhancements/issues/4569
- https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/4569-cgroup-v1-maintenance-mode/README.md KEP-4569: Moving cgroup v1 support into maintenance mode

### DIFF
--- a/validators/cgroup_validator_linux.go
+++ b/validators/cgroup_validator_linux.go
@@ -21,6 +21,7 @@ package system
 
 import (
 	"bufio"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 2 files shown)


validators/cgroup_validator_linux.go
```diff
@@ -68,6 +69,7 @@ func (c *CgroupsValidator) Validate(spec SysSpec) (warns, errs []error) {
 		requiredCgroupSpec = spec.CgroupsV2
 		optionalCgroupSpec = spec.CgroupsV2Optional
 	} else {
+		warns = append(warns, errors.New("cgroup v1 support is in maintenance mode"))
```
neolit123, line 72: cgroup -> cgroups

should we say something like:
```suggestion
		warns = append(warns, errors.New("cgroup v1 support is in maintenance mode. Please upgrade to cgroups v2"))
```
?

pacoxu: Use https://github.com/kubernetes/kubernetes/pull/125328/files#diff-31b906a36201f455ea92866cf7adbd5440cb188fb06515b6e3d01772daec3f8dR47

> Cgroup v1 support is in maintenance mode, please migrate to Cgroup v2.

According to https://man7.org/linux/man-pages/man7/cgroups.7.html, we should use cgroups.

neolit123: yeah `cgroups` is the right name (plural noun with `s`)

neolit123 (commented):
/lgtm
/approve

PR #40, opened 2024-09-06
Windows: do not use 'uname' for doing OS validation

neolit123:
The idea of the OS validator is to check if the host OS is supported. e.g. 'uname' would return 'Linux' on Linux and something else on OSX, BSD.

But 'uname' is a Unix tool, and only available on Windows if third party packages are installed like MSYS.

- Split the os_validator*.go files for Unix and Windows.
- Align the fetching of OS name and kernel version for Windows with the kubelet which uses the CurrentVersion reg key under HKLM.
- In the Windows spec use 'Windows Server' as the required name and match it in 'CurrentVersion.ProductName' as a prefix.
- Add OS specific unit tests.


neolit123:
note: i ran the _windows unit tests locally on WSL
sadly golang still does not support cross-running UT for another OS.

jsturtevant:
ran the tests on my Windows 10 machine:

```
cmd /c ver

Microsoft Windows [Version 10.0.22631.4037]

go test .\validators\
ok      k8s.io/system-validators/validators     1.602s
```

jsturtevant (approved):
/lgtm

pacoxu (commented):
/lgtm

neolit123 (commented):
@pacoxu @jsturtevant 
i pushed a small fix for wrong year in license header
https://github.com/kubernetes/system-validators/compare/f070f228e2386f604341c48b5072c1d0a38b4404..9435f6c43ece976ea9deac83a001108b4846b88e

please re-apply lgtm.

PR #41, opened 2024-09-25
add CONFIG_CGROUP_BPF to required kernel config

### LINKED ISSUE #36: Update the kernel configuration for cgroup v2
Kubernetes 1.25 brings cgroup v2 to GA. 
cgroup v2 needs some additional kernel configs.
For example, you need to enable `CONFIG_CGROUP_BPF` if you want to use the the device controller.
When it comes to Kubernetes, I have never investigated what config is actually needed.
But I just create an issue at first.

pacoxu:
References:
- https://github.com/containerd/containerd/pull/3799#issuecomment-555740694
- https://man7.org/linux/man-pages/man7/cgroups.7.html
- https://www.kernel.org/doc/Documentation/admin-guide/cgroup-v2.rst
- https://www.kernel.org/doc/Documentation/admin-guide/cgroup-v1/

Fixes #36

/cc @KentaTada @neolit123

### DIFF
--- a/validators/cgroup_validator_linux.go
+++ b/validators/cgroup_validator_linux.go
@@ -31,7 +31,10 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-var _ Validator = &CgroupsValidator{}
+var (
+	_                 Validator = &CgroupsValidator{}
+	unifiedMountpoint           = getUnifiedMountpoint()
+)
 
 // CgroupsValidator validates cgroup configuration.
 type CgroupsValidator struct {
@@ -44,9 +47,29 @@ func (c *CgroupsValidator) Name() string {
 }
 
 const (
-	cgroupsConfigPrefix = "CGROUPS_"
-	unifiedMountpoint   = "/sys/fs/cgroup"
+	cgroupsConfigPrefix      = "CGROUPS_"
+	defaultUnifiedMountpoint = "/sys/fs/cgroup"
+	mountsFilePath           = "/proc/mounts"
 )
+
+// getUnifiedMountpoint will check /proc/mounts and return the cgroup dir
+func getUnifiedMountpoint() string {
+	f, err := os.Open(mountsFilePath)
+	if err != nil {
+		fmt.Printf("error checking %q: %v\n", mountsFilePath, err)
+	} else {
+		defer f.Close()
+		scanner := bufio.NewScanner(f)
+		for scanner.Scan() {
+			// example fields: `cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0`
+			fields := strings.Split(scanner.Text(), " ")
+			if len(fields) >= 3 && (fields[2] == "cgroup2" || fields[2] == "cgroup") {
+				return fields[1]
+			}
+		}
+	}
+	return defaultUnifiedMountpoint
+}
 
 // Validate is part of the system.Validator interface.
 func (c *CgroupsValidator) Validate(spec SysSpec) (warns, errs []error) {
@@ -142,15 +165,39 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
 	// Some controllers are implicitly enabled by the kernel.
 	// Those controllers do not appear in /sys/fs/cgroup/cgroup.controllers.
 	// https://github.com/torvalds/linux/blob/v5.3/kernel/cgroup/cgroup.c#L433-L434
-	// We assume these are always available, as it is hard to detect availability.
-	// So, we hardcode the following as "pseudo" controllers.
+	// For freezer, we use checkCgroupV2Freeze() to check.
+	// For others, we assume these are always available, as it is hard to detect availability.
+	// We hardcode the following as initial controllers.
 	// - devices: implemented in kernel 4.15
-	// - freezer: implemented in kernel 5.2
-	pseudo := []string{"devices", "freezer"}
+	subsystems := []string{"devices"}
+	if checkCgroupV2Freeze() {
+		subsystems = append(subsystems, "freezer")
+	}
 	data, err := ioutil.ReadFile(filepath.Join(unifiedMountpoint, "cgroup.controllers"))
 	if err != nil {
 		return nil, err
 	}
-	subsystems := append(pseudo, strings.Fields(string(data))...)
+	subsystems = append(subsystems, strings.Fields(string(data))...)
 	return subsystems, nil
+}
+
+// For freezer which is implemented in kernel 5.2, we can check the existence of `cgroup.freeze`.
+func checkCgroupV2Freeze() bool {
+	tmpDir, err := os.MkdirTemp(unifiedMountpoint, "freezer-test")
+	if err != nil {
+		fmt.Printf("error mkdir under %q: %v\n", unifiedMountpoint, err)
+		return false
+	}
+	defer func() {
+		err := os.RemoveAll(tmpDir)
+		if err != nil {
+			fmt.Printf("error remove dir %q: %v\n", tmpDir, err)
+		}
+	}()
+	_, err = os.Stat(filepath.Join(tmpDir, "/cgroup.freeze"))
+	if err == nil {
+		return true
+	}
+	fmt.Printf("no cgroup.freeze under %q: %v\n", tmpDir, err)
+	return false
 }
--- a/validators/cgroup_validator_linux.go
+++ b/validators/cgroup_validator_linux.go
@@ -62,12 +62,29 @@ func getUnifiedMountpoint() string {
 
 	defer f.Close()
 	scanner := bufio.NewScanner(f)
+	var cgroupv1MountPoint string
 	for scanner.Scan() {
 		// example fields: `cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0`
-		fields := strings.Split(scanner.Text(), " ")
-		if len(fields) >= 3 && (fields[2] == "cgroup2" || fields[2] == "cgroup") {
-			return fields[1]
+		fields := strings.Fields(scanner.Text())
+		if len(fields) >= 3 {
+			switch fields[2] {
+			case "cgroup2":
+				// return the first cgroup v2 mount point directly
+				return fields[1]
+			case "cgroup":
+				// set the first cgroup v1 mount point only,
+				// and continue the loop to find if there is a cgroup v2 mount point
+				if len(cgroupv1MountPoint) == 0 {
+					cgroupv1MountPoint = fields[1]
+				}
+			default:
+				continue
+			}
 		}
+	}
+	// return cgroup v1 mount point if no cgroup v2 mount point is found
+	if len(cgroupv1MountPoint) != 0 {
+		return cgroupv1MountPoint
 	}
 	return defaultUnifiedMountpoint
 }
@@ -199,8 +216,10 @@ func checkCgroupV2Freeze() (bool, error) {
 		}
 	}()
 	_, err = os.Stat(filepath.Join(tmpDir, "cgroup.freeze"))
-	if err == nil {
-		return true, nil
+	if os.IsNotExist(err) {
+		return false, nil
+	} else if err != nil {
+		fmt.Printf("error stat cgroup.freeze file under %q: %v\n", tmpDir, err)
 	}
-	return false, err
+	return true, nil
 }
--- a/validators/cgroup_validator_linux.go
+++ b/validators/cgroup_validator_linux.go
@@ -31,10 +31,7 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-var (
-	_                 Validator = &CgroupsValidator{}
-	unifiedMountpoint           = getUnifiedMountpoint()
-)
+var _ Validator = &CgroupsValidator{}
 
 // CgroupsValidator validates cgroup configuration.
 type CgroupsValidator struct {
@@ -47,65 +44,72 @@ func (c *CgroupsValidator) Name() string {
 }
 
 const (
-	cgroupsConfigPrefix      = "CGROUPS_"
-	defaultUnifiedMountpoint = "/sys/fs/cgroup"
-	mountsFilePath           = "/proc/mounts"
+	cgroupsConfigPrefix = "CGROUPS_"
+	mountsFilePath      = "/proc/mounts"
 )
 
-// getUnifiedMountpoint will check /proc/mounts and return the cgroup dir
-func getUnifiedMountpoint() string {
-	f, err := os.Open(mountsFilePath)
+// getUnifiedMountpoint checks if the default mount point is available.
+// If not, it parses the mounts file to find a valid cgroup mount point.
+func getUnifiedMountpoint(path string) (string, error) {
+	f, err := os.Open(path)
 	if err != nil {
-		fmt.Printf("error checking %q: %v\n", mountsFilePath, err)
-		return defaultUnifiedMountpoint
+		return "", err
 	}
-
 	defer f.Close()
 	scanner := bufio.NewScanner(f)
-	var cgroupv1MountPoint string
+	var cgroupV1MountPoint string
 	for scanner.Scan() {
-		// example fields: `cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0`
-		fields := strings.Fields(scanner.Text())
+		line := scanner.Text()
+		if !strings.Contains(line, "cgroup") {
+			continue
+		}
+		// Example fields: `cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0`.
+		fields := strings.Fields(line)
 		if len(fields) >= 3 {
 			switch fields[2] {
 			case "cgroup2":
-				// return the first cgroup v2 mount point directly
-				return fields[1]
+				// Return the first cgroups v2 mount point directly.
+				return fields[1], nil
 			case "cgroup":
-				// set the first cgroup v1 mount point only,
-				// and continue the loop to find if there is a cgroup v2 mount point
-				if len(cgroupv1MountPoint) == 0 {
-					cgroupv1MountPoint = fields[1]
+				// Set the first cgroups v1 mount point only,
+				// and continue the loop to find if there is a cgroups v2 mount point.
+				if len(cgroupV1MountPoint) == 0 {
+					cgroupV1MountPoint = fields[1]
 				}
-			default:
-				continue
 			}
 		}
 	}
-	// return cgroup v1 mount point if no cgroup v2 mount point is found
-	if len(cgroupv1MountPoint) != 0 {
-		return cgroupv1MountPoint
+	// Return cgroups v1 mount point if no cgroups v2 mount point is found.
+	if len(cgroupV1MountPoint) != 0 {
+		return cgroupV1MountPoint, nil
 	}
-	return defaultUnifiedMountpoint
+	return "", fmt.Errorf("cannot get a cgroupfs mount point from %q", path)
 }
 
 // Validate is part of the system.Validator interface.
 func (c *CgroupsValidator) Validate(spec SysSpec) (warns, errs []error) {
-	// Get the subsystems from /sys/fs/cgroup/cgroup.controllers when cgroup v2 is used.
+	// Get the subsystems from /sys/fs/cgroup/cgroup.controllers when cgroups v2 is used.
 	// /proc/cgroups is meaningless for v2
 	// https://github.com/torvalds/linux/blob/v5.3/Documentation/admin-guide/cgroup-v2.rst#deprecated-v1-core-features
 	var st unix.Statfs_t
-	var err error
+	unifiedMountpoint, err := getUnifiedMountpoint(mountsFilePath)
+	if err != nil {
+		return nil, []error{fmt.Errorf("cannot get a cgroup mount point: %w", err)}
+	}
 	if err := unix.Statfs(unifiedMountpoint, &st); err != nil {
 		return nil, []error{fmt.Errorf("cannot statfs the cgroupv2 root: %w", err)}
 	}
 	var requiredCgroupSpec []string
 	var optionalCgroupSpec []string
 	var subsystems []string
+	var warn error
 	if st.Type == unix.CGROUP2_SUPER_MAGIC {
-		subsystems, err = c.getCgroupV2Subsystems()
+		subsystems, err, warn = c.getCgroupV2Subsystems(unifiedMountpoint)
 		if err != nil {
 			return nil, []error{fmt.Errorf("failed to get cgroups v2 subsystems: %w", err)}
+		}
+		if warn != nil {
+			warns = append(warns, warn)
 		}
 		requiredCgroupSpec = spec.CgroupsV2
 		optionalCgroupSpec = spec.CgroupsV2Optional
@@ -151,11 +155,10 @@ func (c *CgroupsValidator) validateCgroupSubsystems(cgroups, subsystems []string
 		missing = append(missing, cgroup)
 	}
 	return missing
-
 }
 
 func (c *CgroupsValidator) getCgroupV1Subsystems() ([]string, error) {
-	// Get the subsystems from /proc/cgroups when cgroup v1 is used.
+	// Get the subsystems from /proc/cgroups when cgroups v1 is used.
 	f, err := os.Open("/proc/cgroups")
 	if err != nil {
 		return nil, err
@@ -179,47 +182,50 @@ func (c *CgroupsValidator) getCgroupV1Subsystems() ([]string, error) {
 	return subsystems, nil
 }
 
-func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
+func (c *CgroupsValidator) getCgroupV2Subsystems(unifiedMountpoint string) ([]string, error, error) {
 	// Some controllers are implicitly enabled by the kernel.
 	// Those controllers do not appear in /sys/fs/cgroup/cgroup.controllers.
 	// https://github.com/torvalds/linux/blob/v5.3/kernel/cgroup/cgroup.c#L433-L434
 	// For freezer, we use checkCgroupV2Freeze() to check.
 	// For others, we assume these are always available, as it is hard to detect availability.
 	// We hardcode the following as initial controllers.
-	// - devices: implemented in kernel 4.15
+	// - devices: implemented in kernel 4.15.
 	subsystems := []string{"devices"}
-	freezeSupported, err := checkCgroupV2Freeze()
-	if err != nil {
-		fmt.Printf("error checking cgroups v2 freeze: %v\n", err)
-	}
+	freezeSupported, warn := checkCgroupV2Freeze(unifiedMountpoint)
 	if freezeSupported {
 		subsystems = append(subsystems, "freezer")
 	}
 	data, err := ioutil.ReadFile(filepath.Join(unifiedMountpoint, "cgroup.controllers"))
 	if err != nil {
-		return nil, err
+		return nil, err, warn
 	}
 	subsystems = append(subsystems, strings.Fields(string(data))...)
-	return subsystems, nil
+	return subsystems, err, warn
 }
 
-// To check if the freezer controller is enabled in Linux kernels 5.2, we can check the existence of cgroup.freeze.
-func checkCgroupV2Freeze() (bool, error) {
-	tmpDir, err := os.MkdirTemp(unifiedMountpoint, "freezer-test")
-	if err != nil {
-		return false, err
+// checkCgroupV2Freeze checks if the freezer controller is enabled in Linux kernels 5.2.
+// It determines that by creating a cgroup.freeze file under the unified mountpoint location.
+func checkCgroupV2Freeze(unifiedMountpoint string) (isCgroupfs bool, warn error) {
+	const freezeFile = "cgroup.freeze"
+	tmpDir, warn := os.MkdirTemp(unifiedMountpoint, "freezer-test")
+	if warn != nil {
+		return
 	}
 	defer func() {
 		err := os.RemoveAll(tmpDir)
 		if err != nil {
-			fmt.Printf("error removing directory %q: %v\n", tmpDir, err)
+			warn = fmt.Errorf("error removing directory %q: %v", tmpDir, err)
 		}
 	}()
-	_, err = os.Stat(filepath.Join(tmpDir, "cgroup.freeze"))
-	if os.IsNotExist(err) {
-		return false, nil
-	} else if err != nil {
-		fmt.Printf("error stat cgroup.freeze file under %q: %v\n", tmpDir, err)
+	_, warn = os.Stat(filepath.Join(tmpDir, freezeFile))
+	if os.IsNotExist(warn) {
+		return
+	} else if warn != nil {
+		// If the err is not NotExist error, it means that `cgroup.freeze` exists.
+		isCgroupfs = true
+		warn = fmt.Errorf("could not stat %q file in %q: %v", freezeFile, tmpDir, warn)
+		return
 	}
-	return true, nil
+	isCgroupfs = true
+	return
 }
--- a/validators/cgroup_validator_linux.go
+++ b/validators/cgroup_validator_linux.go
@@ -57,15 +57,16 @@ func getUnifiedMountpoint() string {
 	f, err := os.Open(mountsFilePath)
 	if err != nil {
 		fmt.Printf("error checking %q: %v\n", mountsFilePath, err)
-	} else {
-		defer f.Close()
-		scanner := bufio.NewScanner(f)
-		for scanner.Scan() {
-			// example fields: `cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0`
-			fields := strings.Split(scanner.Text(), " ")
-			if len(fields) >= 3 && (fields[2] == "cgroup2" || fields[2] == "cgroup") {
-				return fields[1]
-			}
+		return defaultUnifiedMountpoint
+	}
+
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		// example fields: `cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0`
+		fields := strings.Split(scanner.Text(), " ")
+		if len(fields) >= 3 && (fields[2] == "cgroup2" || fields[2] == "cgroup") {
+			return fields[1]
 		}
 	}
 	return defaultUnifiedMountpoint
@@ -170,7 +171,11 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
 	// We hardcode the following as initial controllers.
 	// - devices: implemented in kernel 4.15
 	subsystems := []string{"devices"}
-	if checkCgroupV2Freeze() {
+	freezeSupported, err := checkCgroupV2Freeze()
+	if err != nil {
+		fmt.Printf("error checking cgroups v2 freeze: %v\n", err)
+	}
+	if freezeSupported {
 		subsystems = append(subsystems, "freezer")
 	}
 	data, err := ioutil.ReadFile(filepath.Join(unifiedMountpoint, "cgroup.controllers"))
@@ -181,23 +186,21 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
 	return subsystems, nil
 }
 
-// For freezer which is implemented in kernel 5.2, we can check the existence of `cgroup.freeze`.
-func checkCgroupV2Freeze() bool {
+// To check if the freezer controller is enabled in Linux kernels 5.2, we can check the existence of cgroup.freeze.
+func checkCgroupV2Freeze() (bool, error) {
 	tmpDir, err := os.MkdirTemp(unifiedMountpoint, "freezer-test")
 	if err != nil {
-		fmt.Printf("error mkdir under %q: %v\n", unifiedMountpoint, err)
-		return false
+		return false, err
 	}
 	defer func() {
 		err := os.RemoveAll(tmpDir)
 		if err != nil {
-			fmt.Printf("error remove dir %q: %v\n", tmpDir, err)
+			fmt.Printf("error removing directory %q: %v\n", tmpDir, err)
 		}
 	}()
-	_, err = os.Stat(filepath.Join(tmpDir, "/cgroup.freeze"))
+	_, err = os.Stat(filepath.Join(tmpDir, "cgroup.freeze"))
 	if err == nil {
-		return true
+		return true, nil
 	}
-	fmt.Printf("no cgroup.freeze under %q: %v\n", tmpDir, err)
-	return false
+	return false, err
 }
--- a/validators/cgroup_validator_test.go
+++ b/validators/cgroup_validator_test.go
@@ -20,13 +20,14 @@ limitations under the License.
 package system
 
 import (
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
 func TestValidateCgroupSubsystem(t *testing.T) {
-	// hardcoded cgroup v2 subsystems
+	// hardcoded cgroups v2 subsystems
 	pseudoSubsystems := []string{"devices", "freezer"}
 
 	v := &CgroupsValidator{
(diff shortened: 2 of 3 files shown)


pacoxu:
As cgroup v1 is in maintenance mode but not removed, we should not remove the cgroup v1 related kernel config out.

However, should we show different message for `CONFIG_CGROUP_BPF` missing?
Or should we check if the user is in cgroup v1 or v2 at this point to decide which check policy to use?

validators/types_unix.go
```diff
@@ -40,11 +40,12 @@ var DefaultSysSpec = SysSpec{
 			{Name: "IPC_NS"},
 			{Name: "UTS_NS"},
 			{Name: "CGROUPS"},
-			{Name: "CGROUP_CPUACCT"},
+			{Name: "CGROUP_BPF"},     // cgroups v2
```
neolit123, line 43: do you know where does it say that v2 requires CGROUP_BPF?
no info here:
https://docs.kernel.org/admin-guide/cgroup-v2.html

neolit123 (commented):
in https://github.com/kubernetes/system-validators/issues/36 @KentaTada mentioned that we need to read another file to verify cgroupv2.

> Unlike in v1, the kernel config alone cannot determine whether the required v2 features are enabled.

validators/types_unix.go
pacoxu: `device` controller is based on BPF. Let me check if there is more official information about that.

pacoxu:
> in #36 @KentaTada mentioned that we need to read another file to verify cgroupv2.

It seems that we need to check in different ways for cgroup v1 and cgroup v2.

For instance,
- cgroup v2: check `/sys/fs/cgroup/init.scope/cgroup.freeze`
- cgroup v1: check CONFIG_CGROUP_FREEZER.

/hold

pacoxu:
```
	// /proc/cgroups is meaningless for v2
	// https://github.com/torvalds/linux/blob/v5.3/Documentation/admin-guide/cgroup-v2.rst#deprecated-v1-core-features
	if IsCgroup2UnifiedMode() {
		// "pseudo" controllers do not appear in /sys/fs/cgroup/cgroup.controllers.
		// - devices: implemented in kernel 4.15
		// - freezer: implemented in kernel 5.2
		// We assume these are always available, as it is hard to detect availability.
		pseudo := []string{"devices", "freezer"}
		data, err := ReadFile("/sys/fs/cgroup", "cgroup.controllers")
		if err != nil {
			return nil, err
		}
		subsystems := append(pseudo, strings.Fields(data)...)
		return subsystems, nil
	}
```
Runc code  https://github.com/kubernetes/kubernetes/blob/e404a28fae6946483c5de0e9f4b4adf6e6f63f3b/vendor/github.com/opencontainers/runc/libcontainer/cgroups/utils.go#L100-L114

pacoxu:
https://github.com/kubernetes/system-validators/blob/c4cf63549977e8d75add5bb76a1be4acfc727793/validators/cgroup_validator_linux.go#L141-L156

```
	pseudo := []string{"devices", "freezer"}
```
We currently use the similar logic here and add `freezer` as default.

https://github.com/kubernetes/system-validators/blob/c4cf63549977e8d75add5bb76a1be4acfc727793/validators/types_unix.go#L69-L70

This checks the cgroup sub controllers.

validators/types_unix.go
pacoxu:
```
Cgroup v2 device controller has no interface files and is implemented
on top of cgroup BPF. 
```
From https://www.kernel.org/doc/Documentation/cgroup-v2.txt.

neolit123:
aligning with RUNC seems good to me. 

what about these comments about freeze vs freezer:
https://github.com/kubernetes/system-validators/issues/36#issuecomment-2285262696

pacoxu:
/unhold
For cgroup v1 cleanup, we should not do it now.

> what about these comments about freeze vs freezer?
> [#36 (comment)](https://github.com/kubernetes/system-validators/issues/36#issuecomment-2285262696)

Updated.

neolit123 (commented):
/lgtm
/approve
/hold

@KentaTada could you please have a look too?

neolit123:
> For cgroup v1 cleanup, we should not do it now.

could you explain why?

pacoxu:
> > For cgroup v1 cleanup, we should not do it now.
> 
> could you explain why?

Basically two reasons:
1. Kubernetes 1.31: Moving cgroup v1 Support into Maintenance Mode, not deprecated.
2. Most OS distrbutions keeps the cgroup v1 related kernel config as is, and I think this may be for backward compatibility.

neolit123:
> > > For cgroup v1 cleanup, we should not do it now.
> > 
> > 
> > could you explain why?
> 
> Basically two reasons IMO:
> 
> 1. Kubernetes 1.31: Moving cgroup v1 Support into Maintenance Mode, not deprecated.
> 2. Most OS distrbutions keeps the cgroup v1 related kernel config as is, and I think this may be for backward compatibility.

ok, got it. you mean that we should not cleanup cgroups v1 because they are still widely used.

i tried finding @KentaTada 's slack to give him a ping but couldn't.
let's wait until end of next week maybe and if he does not reply we can go merge this PR as it is.

KentaTada:
> @KentaTada could you please have a look too?

I'll take a look at this pr by the first half of the next week.

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +142,19 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
-	// We assume these are always available, as it is hard to detect availability.
-	// So, we hardcode the following as "pseudo" controllers.
+	// For freezer which is implemented in kernel 5.2, we can check the existence of `cgroup.freeze`.
+	// For others, we assume these are always available, as it is hard to detect availability.
+	// We hardcode the following as initial controllers.
 	// - devices: implemented in kernel 4.15
-	// - freezer: implemented in kernel 5.2
-	pseudo := []string{"devices", "freezer"}
+	subsystems := []string{"devices"}
+	_, err := os.Stat("/sys/fs/cgroup/init.scope/cgroup.freeze")
```
KentaTada, line 150: It might be a corner case, but we want to get a path of cgroup.freeze
in a way that is independent of the environment.
- "/sys/fs/cgroup" is the default mountpoint and modifiable.
- "init.scope" is specific for systemd.

validators/types_unix.go
```diff
@@ -40,11 +40,12 @@ var DefaultSysSpec = SysSpec{
...
+			{Name: "CGROUP_CPUACCT"}, // cgroups v1 cpuacct
 			{Name: "CGROUP_DEVICE"},
 			{Name: "CGROUP_FREEZER"},
 			{Name: "CGROUP_PIDS"},
-			{Name: "CGROUP_SCHED"},
+			{Name: "CGROUP_SCHED"}, // cgroups v1 cpu
```
KentaTada, line 48: It seems that Utilization Clamping depends on CONFIG_CGROUP_SCHED.
https://github.com/torvalds/linux/blob/v6.12-rc1/init/Kconfig#L1080
Since the values of Utilization Clamping can be set through the cgroup
v2 interface, CONFIG_CGROUP_SCHED must also be enabled for cgroup v2.
https://github.com/torvalds/linux/blob/v6.12-rc1/kernel/sched/core.c#L9920

validators/types_unix.go
```diff
@@ -40,11 +40,12 @@ var DefaultSysSpec = SysSpec{
 			{Name: "IPC_NS"},
 			{Name: "UTS_NS"},
 			{Name: "CGROUPS"},
-			{Name: "CGROUP_CPUACCT"},
+			{Name: "CGROUP_BPF"},     // cgroups v2
+			{Name: "CGROUP_CPUACCT"}, // cgroups v1 cpuacct
 			{Name: "CGROUP_DEVICE"},
 			{Name: "CGROUP_FREEZER"},
```
KentaTada, line 46: CONFIG_CGROUP_FREEZER is for cgroups v1 freezer.
According to the below, we may disable CONFIG_CGROUP_FREEZER when we
use cgroups v2 freezer.
https://github.com/torvalds/linux/blob/v6.12-rc1/init/Kconfig#L1132

KentaTada:
@pacoxu @neolit123 
I'm sorry for the late reply.
I left some comments.
In addition to that, I think we should split the structure of the required kernel config into two parts: one for v1 and the other for v2?
If users are using v2, they won’t want to see the list for v1.

validators/cgroup_validator_linux.go
pacoxu: > * "/sys/fs/cgroup" is the default mountpoint and modifiable.

https://github.com/kubernetes/system-validators/blob/1d2f5889f07347b9f2d3ed337b3cfb08d45f4cf6/validators/cgroup_validator_linux.go#L48

We currently use this default here for other checks.

If not, we may have to check the `mount`  or `/proc/mounts`/`/proc/self/mounts` to get the cgroup2 mount path. 


> "init.scope" is specific for systemd.

Or we may use `system.slice/cgroup.freeze`?

validators/types_unix.go
pacoxu: Updated.

validators/cgroup_validator_linux.go
pacoxu:
```
	_, err := os.Stat(filepath.Join(unifiedMountpoint, "system.slice", "/cgroup.freeze"))
```
Changed to this. 

https://github.com/kubernetes/system-validators/blob/4d61a4e8f440a37c88e281ae115e43e8893683a1/validators/cgroup_validator_linux.go#L150-L154

pacoxu:
BTW, I am trying to write a blog about this in https://github.com/kubernetes/website/pull/47342/. Could you help to review as well？

>> In addition to that, I think we should split the structure of the required kernel config into two parts: one for v1 and the other for v2?
>> If users are using v2, they won’t want to see the list for v1.
> IIUC, the only way to split the v1 and v2 spec is to introduce a breaking change in the spec API, which means we need to release a new MAJOR version of this small library.

If so, may we do  that later?

neolit123:
there seem to be code org benefits from it, but overall, as long as we add the proper comment in the spec we can avoid changing the spec API?

i don't know how many users this lib has, but likely there are not many.
either way, we should follow semver rules for it.

pacoxu:
I would proposal to do this in comment now for the current release. 
And later, we can refactor and release a new MAJOR version. Do we need a tracking issue or major release branch for that(it seems that we don't have other new features here)?

neolit123:
> I would proposal to do this in comment now for the current release. And later, we can refactor and release a new MAJOR version. Do we need a tracking issue or major release branch for that(it seems that we don't have other new features here)?

i don't think we need a tracking issue.
also, thinking more about it we should schedule this breaking change with another one. i.e. don't release a MAJOR version only because of the cgroups API refactor.

validators/cgroup_validator_linux.go
KentaTada: Thank you for your suggestion.

> If not, we may have to check the `mount` or `/proc/mounts`/`/proc/self/mounts` to get the cgroup2 mount path.

Yes. I think we need to get the path from /proc/mounts.
@neolit123  Is this change ok? It may affect the behavior of existing check.

> Or we may use system.slice/cgroup.freeze?

system.slice is also dependent on systemd.
Creating a directory in cgroupfs for this check is the perfect solution.

KentaTada:
> i don't think we need a tracking issue.
> also, thinking more about it we should schedule this breaking change with another one. i.e. don't release a MAJOR version only because of the cgroups API refactor.

OK. I agree with you.
Because don't need to split the v1 and v2 spec, I think we should just leave a suitable comment.

KentaTada:
> BTW, I am trying to write a blog about this in [kubernetes/website#47342](https://github.com/kubernetes/website/pull/47342). Could you help to review as well？

Yes!
I'd like to review it.
Thanks!

validators/cgroup_validator_linux.go
neolit123: unfortunately i don't know a lot about the details here.

one point to make - kubeadm supports the openrc init system from alpine and not only systemd:
https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/util/initsystem/initsystem_unix.go#L46C14-L46C20

also, k8s kubelet currently only works with the systemd cgroups driver AFAIK.
but for runc it's not mandatory:
https://github.com/opencontainers/runc/blob/main/docs/cgroup-v2.md#systemd 

> system.slice is also dependent on systemd.

yes there are still users using the cgroupfs driver, which means that system.slice will not exist.

> Creating a directory in cgroupfs for this check is the perfect solution.

could you elaborate why we need to create a dir?

IIUC what we need to do to check if a controller foo is enabled is to do the following:
```
check if these paths exist:
  - for cgroupfs /sys/fs/cgroup/freezer
   - for systemd /sys/fs/cgroup/systemd/freezer.state
```
if one of them exists we can say freezer is enabled, i don't think we need to check freezer.state for values.

pacoxu: Updated.

You can see https://github.com/kubernetes/system-validators/compare/4d61a4e8f440a37c88e281ae115e43e8893683a1..2ec937dbfb62d8fa673a16e73db0fa21eb042d3b. 

@KentaTada Do you have time to take another look?

neolit123: related
- https://github.com/k0sproject/k0s/issues/3069

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +165,39 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
+		fmt.Printf("error mkdir under %q: %v\n", unifiedMountpoint, err)
+		return false
+	}
+	defer func() {
+		err := os.RemoveAll(tmpDir)
+		if err != nil {
+			fmt.Printf("error remove dir %q: %v\n", tmpDir, err)
+		}
+	}()
+	_, err = os.Stat(filepath.Join(tmpDir, "/cgroup.freeze"))
```
neolit123, line 198:
```suggestion
	_, err = os.Stat(filepath.Join(tmpDir, "cgroup.freeze"))
```

is that additional / needed?

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +165,39 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
 	if err != nil {
 		return nil, err
 	}
-	subsystems := append(pseudo, strings.Fields(string(data))...)
+	subsystems = append(subsystems, strings.Fields(string(data))...)
 	return subsystems, nil
 }
+
+// For freezer which is implemented in kernel 5.2, we can check the existence of `cgroup.freeze`.
+func checkCgroupV2Freeze() bool {
```
neolit123, line 186:
```suggestion
func checkCgroupV2Freeze() (error, bool) {
```

instead of print return error and bool

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +165,39 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
+	tmpDir, err := os.MkdirTemp(unifiedMountpoint, "freezer-test")
+	if err != nil {
+		fmt.Printf("error mkdir under %q: %v\n", unifiedMountpoint, err)
+		return false
+	}
+	defer func() {
+		err := os.RemoveAll(tmpDir)
+		if err != nil {
+			fmt.Printf("error remove dir %q: %v\n", tmpDir, err)
```
neolit123, line 195:
```suggestion
			fmt.Printf("error removing directory %q: %v\n", tmpDir, err)
```

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +165,39 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
-	subsystems := append(pseudo, strings.Fields(string(data))...)
+	subsystems = append(subsystems, strings.Fields(string(data))...)
 	return subsystems, nil
 }
+
+// For freezer which is implemented in kernel 5.2, we can check the existence of `cgroup.freeze`.
+func checkCgroupV2Freeze() bool {
+	tmpDir, err := os.MkdirTemp(unifiedMountpoint, "freezer-test")
+	if err != nil {
+		fmt.Printf("error mkdir under %q: %v\n", unifiedMountpoint, err)
```
neolit123, line 189:
```suggestion
		fmt.Printf("error crearing directory under %q: %v\n", unifiedMountpoint, err)
```

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +165,39 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
 	data, err := ioutil.ReadFile(filepath.Join(unifiedMountpoint, "cgroup.controllers"))
 	if err != nil {
 		return nil, err
 	}
-	subsystems := append(pseudo, strings.Fields(string(data))...)
+	subsystems = append(subsystems, strings.Fields(string(data))...)
 	return subsystems, nil
 }
+
+// For freezer which is implemented in kernel 5.2, we can check the existence of `cgroup.freeze`.
```
neolit123, line 185: // To check if the freezer controller is enabled in Linux kernels 5.2, we can check the existence of `cgroup.freeze`.

neolit123 (commented):
i don't know if creating the tmp dir is the right solution.
leaving review to @KentaTada

validators/cgroup_validator_linux.go
KentaTada: > could you elaborate why we need to create a dir?

/sys/fs/cgroup/freezer does not exist.
Like k0s, we also need to create a dummy cgroup:
https://github.com/k0sproject/k0s/pull/3077/files#diff-468315026edb0a5cee0863f23231ad03809dcf5aa33958e7df1fd03779aa2621R171

validators/cgroup_validator_linux.go
```diff
@@ -44,10 +47,31 @@ func (c *CgroupsValidator) Name() string {
...
-	unifiedMountpoint   = "/sys/fs/cgroup"
+	cgroupsConfigPrefix      = "CGROUPS_"
+	defaultUnifiedMountpoint = "/sys/fs/cgroup"
+	mountsFilePath           = "/proc/mounts"
 )
 
+// getUnifiedMountpoint will check /proc/mounts and return the cgroup dir
+func getUnifiedMountpoint() string {
+	f, err := os.Open(mountsFilePath)
+	if err != nil {
```
KentaTada, line 58: fstype should be checked to confirm whether defaultUnifiedMountpoint is actually cgroupfs.

validators/cgroup_validator_linux.go
```diff
@@ -44,10 +47,31 @@ func (c *CgroupsValidator) Name() string {
...
+		fmt.Printf("error checking %q: %v\n", mountsFilePath, err)
+		return defaultUnifiedMountpoint
+	}
+
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		// example fields: `cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0`
+		fields := strings.Split(scanner.Text(), " ")
```
KentaTada, line 67: Just in case, I think that strings.Fields is suitable to handle spaces like below:
https://github.com/kubernetes/system-validators/blob/1d2f5889f07347b9f2d3ed337b3cfb08d45f4cf6/validators/cgroup_validator_linux.go#L132

validators/cgroup_validator_linux.go
```diff
@@ -44,10 +47,31 @@ func (c *CgroupsValidator) Name() string {
...
+		if len(fields) >= 3 && (fields[2] == "cgroup2" || fields[2] == "cgroup") {
```
KentaTada, line 68: Which shoud we select the mount point if there are multiple cgroup mount points?

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +166,41 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
+		return false, err
+	}
+	defer func() {
+		err := os.RemoveAll(tmpDir)
+		if err != nil {
+			fmt.Printf("error removing directory %q: %v\n", tmpDir, err)
+		}
+	}()
+	_, err = os.Stat(filepath.Join(tmpDir, "cgroup.freeze"))
+	if err == nil {
```
KentaTada, line 202: Can we use os.IsNotExist(err) to clarify the root cause?
Not the absence of the cgroup.freeze but other errors may happen.

validators/cgroup_validator_linux.go
KentaTada: > @KentaTada Do you have time to take another look?

I took another look.
Could you confirm it?

validators/cgroup_validator_linux.go
pacoxu: The current logic is to return the first `cgroup` or `cgroup2`.

Do you mean that we can return `cgroup2` if we have? And if not, we return the `cgroup` mount point.

validators/cgroup_validator_linux.go
pacoxu: if we cannot open the `/proc/mounts`(err!=nil here), how can we confirm about the default mountpoint groupfs?

- check `/proc/self/mountinfo`?

validators/cgroup_validator_linux.go
pacoxu: Updated.

validators/cgroup_validator_linux.go
pacoxu: Updated.

validators/cgroup_validator_linux.go
pacoxu: Updated.
- Should I print error if it is not `NotExistErr`?

pacoxu:
I updated the PR accordingly and the only unresolved comment is https://github.com/kubernetes/system-validators/pull/41/files#r1800338762. 
- I am not sure how to confirm about the default mount point if we cannot get the mount info correctly.

validators/cgroup_validator_linux.go
pacoxu: https://github.com/kubernetes/system-validators/pull/41/commits/95629f0e852ddc8c74bdb25c98cf144509146018 tried to address this.

 1. check /proc/self/mountinfo if the default mount point is available
 2. check /proc/mounts and return the cgroup dir

pacoxu: @neolit123 @KentaTada  PTAL

validators/cgroup_validator_linux.go
```diff
@@ -44,10 +47,96 @@ func (c *CgroupsValidator) Name() string {
...
-	unifiedMountpoint   = "/sys/fs/cgroup"
+	cgroupsConfigPrefix      = "CGROUPS_"
+	defaultUnifiedMountpoint = "/sys/fs/cgroup"
+	mountsFilePath           = "/proc/mounts"
+	mountInfoPath            = "/proc/self/mountinfo"
 )
 
+// 1. check /proc/self/mountinfo if the default mount point is available
+// 2. check /proc/mounts and return the cgroup dir
+func getUnifiedMountpoint() string {
```
neolit123, line 58: make this function return a string, error instead of calling printf() in it.

neolit123, line 58: make this function return a string, error instead of calling printf() in it.

validators/cgroup_validator_linux.go
```diff
@@ -44,10 +47,96 @@ func (c *CgroupsValidator) Name() string {
...
 const (
-	cgroupsConfigPrefix = "CGROUPS_"
-	unifiedMountpoint   = "/sys/fs/cgroup"
+	cgroupsConfigPrefix      = "CGROUPS_"
+	defaultUnifiedMountpoint = "/sys/fs/cgroup"
+	mountsFilePath           = "/proc/mounts"
+	mountInfoPath            = "/proc/self/mountinfo"
 )
 
+// 1. check /proc/self/mountinfo if the default mount point is available
```
neolit123, line 56:
```suggestion
// getUnifiedMountpoint does the following:
// 1. check /proc/self/mountinfo if the default mount point is available
```

validators/cgroup_validator_linux.go
```diff
@@ -44,10 +47,96 @@ func (c *CgroupsValidator) Name() string {
...
+				}
+			}
+		}
+	}
+	// return cgroup v1 mount point if no cgroup v2 mount point is found
+	if len(cgroupv1MountPoint) != 0 {
+		return cgroupv1MountPoint, nil
+	}
+	return "", fmt.Errorf("cannot get a cgroupfs mount point from %q", path)
+
```
neolit123, line 137:
```suggestion
```

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +231,43 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
+	// For freezer, we use checkCgroupV2Freeze() to check.
+	// For others, we assume these are always available, as it is hard to detect availability.
+	// We hardcode the following as initial controllers.
 	// - devices: implemented in kernel 4.15
-	// - freezer: implemented in kernel 5.2
-	pseudo := []string{"devices", "freezer"}
+	subsystems := []string{"devices"}
+	freezeSupported, err := checkCgroupV2Freeze()
+	if err != nil {
+		fmt.Printf("error checking cgroups v2 freeze: %v\n", err)
```
neolit123, line 241: instead of printing here i think the function should return []string, error, error
```
		subsystems, warn, err = c.getCgroupV2Subsystems()
```

the caller function has it:
```
func (c *CgroupsValidator) Validate(spec SysSpec) (warns, errs []error) {
```

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +231,43 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
 	if err != nil {
 		return nil, err
 	}
-	subsystems := append(pseudo, strings.Fields(string(data))...)
+	subsystems = append(subsystems, strings.Fields(string(data))...)
 	return subsystems, nil
 }
+
+// To check if the freezer controller is enabled in Linux kernels 5.2, we can check the existence of cgroup.freeze.
+func checkCgroupV2Freeze() (bool, error) {
```
neolit123, line 255:
```suggestion
// checkCgroupV2Freeze checks if the freezer controller is enabled in Linux kernels 5.2, we can check the existence of cgroup.freeze.
func checkCgroupV2Freeze() (bool, error) {
```

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +231,43 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
+	tmpDir, err := os.MkdirTemp(unifiedMountpoint, "freezer-test")
+	if err != nil {
+		return false, err
+	}
+	defer func() {
+		err := os.RemoveAll(tmpDir)
+		if err != nil {
+			fmt.Printf("error removing directory %q: %v\n", tmpDir, err)
```
neolit123, line 263: return warning or error don't print.

validators/cgroup_validator_linux.go
```diff
@@ -142,15 +231,43 @@ func (c *CgroupsValidator) getCgroupV2Subsystems() ([]string, error) {
...
+		}
+	}()
+	_, err = os.Stat(filepath.Join(tmpDir, "cgroup.freeze"))
+	if os.IsNotExist(err) {
+		return false, nil
+	} else if err != nil {
+		fmt.Printf("error stat cgroup.freeze file under %q: %v\n", tmpDir, err)
```
neolit123, line 270: same here

validators/cgroup_validator_linux.go
pacoxu: Do we need this warning? 
https://github.com/kubernetes/system-validators/pull/41/files#diff-0cb89d807d0cf7f118534018122b93367efce21da4e1650705f6c991872ae3e4R63

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,116 @@ func (c *CgroupsValidator) Name() string {
...
 )
 
+// getUnifiedMountpoint checks if the default mount point is available.
+// If not, it parses the mounts file to find a valid cgroup mount point.
+func getUnifiedMountpoint() (string, error) {
+	isCgroupfs, err := isMountPointCgroupfs(defaultUnifiedMountpoint)
+	if isCgroupfs {
+		return defaultUnifiedMountpoint, err
+	}
+	fmt.Printf("warning: %q is not cgroupfs: %v\n", defaultUnifiedMountpoint, err)
```
pacoxu, line 60: This is the only left `Printf`.
Is this warning needed?  Or we just ignore it.

validators/cgroup_validator_linux.go
neolit123: i think the url you posted is not correct ^
but basically if this func returns err the caller can decide if it's a warning or error.

pacoxu: https://github.com/kubernetes/system-validators/pull/41/files#r1800594545 This link.

neolit123: print should only be done from the main Validate() function.
let's try to show this warning on the top caller.

once you update i will check again.

pacoxu: return it as a warning.

Updated.

KentaTada:
@pacoxu 
Thank you for improving this feature.
It looks good to me.

pacoxu:
> @pacoxu
> Thank you for improving this feature.
> It looks good to me.

Thanks for your detailed review and comments. 
Wait for @neolit123 to double check if anything is missing.

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,117 @@ func (c *CgroupsValidator) Name() string {
...
+	mountInfoPath            = "/proc/self/mountinfo"
 )
 
+// getUnifiedMountpoint checks if the default mount point is available.
+// If not, it parses the mounts file to find a valid cgroup mount point.
+func getUnifiedMountpoint() (string, error, error) {
+	isCgroupfs, err := isMountPointCgroupfs(defaultUnifiedMountpoint)
+	if isCgroupfs {
+		return defaultUnifiedMountpoint, err, nil
+	}
```
neolit123, line 59:
```suggestion
	isCgroupfs, err := isMountPointCgroupfs(defaultUnifiedMountpoint)
	if err != nil {
		return "", err, nil
	}
	if isCgroupfs {
		return defaultUnifiedMountpoint, nil, nil
	}
```
otherwise the err returned by isMountPointCgroupfs is never processed.

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,117 @@ func (c *CgroupsValidator) Name() string {
...
+	mountPath, err := parseMountsFile(mountsFilePath)
+	return mountPath, err, fmt.Errorf("%q is not cgroupfs: %v", defaultUnifiedMountpoint, err)
```
neolit123, line 61:
```suggestion
	return mountPath, nil, warn
```

warn can be constructed earlier.

should it be here?

```
	isCgroupfs, err := isMountPointCgroupfs(defaultUnifiedMountpoint)
	if err != nil {
		warn := fmt.Errorf("%q is not cgroupfs: %v", defaultUnifiedMountpoint, err)
	}
```

i got confused what was the err and what was the warn

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,117 @@ func (c *CgroupsValidator) Name() string {
...
 )
 
+// getUnifiedMountpoint checks if the default mount point is available.
+// If not, it parses the mounts file to find a valid cgroup mount point.
+func getUnifiedMountpoint() (string, error, error) {
+	isCgroupfs, err := isMountPointCgroupfs(defaultUnifiedMountpoint)
+	if isCgroupfs {
+		return defaultUnifiedMountpoint, err, nil
+	}
+	mountPath, err := parseMountsFile(mountsFilePath)
```
neolit123, line 60:
```suggestion
	mountPath, err := parseMountsFile(mountsFilePath)
	if err != nil {
		return "", err, warn
	}
```

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,117 @@ func (c *CgroupsValidator) Name() string {
...
+	return mountPath, err, fmt.Errorf("%q is not cgroupfs: %v", defaultUnifiedMountpoint, err)
+}
+
+func isMountPointCgroupfs(path string) (bool, error) {
```
neolit123, line 64: i think this should be unit tested.

you can break it down into two functions:
isMountPointCgroupfs opens that file and reads the []byte / string with isMountPointCgroupfsImpl()

impl == implementation. we it this way in some places in kubeadm.
isMountPointCgroupfsImpl should be unit tested.

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,117 @@ func (c *CgroupsValidator) Name() string {
...
+	}
+
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if !strings.Contains(line, "cgroup") {
+			continue
+		}
+		// example fields: `28 22 0:25 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:4 - cgroup2 cgroup2 rw,seclabel`
```
neolit123, line 77: the format should be consistent across kernels and distros, is that true?

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,117 @@ func (c *CgroupsValidator) Name() string {
...
+		fields := strings.Fields(line)
+		if len(fields) >= 9 {
```
neolit123, line 79: what does it mean if there are less fields on the line?

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,117 @@ func (c *CgroupsValidator) Name() string {
...
+			if fields[5] == path {
+				i := 6
+				for i < len(fields) && fields[i] != "-" {
+					i++
```
neolit123, line 83: can this iterator be avoided?

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,117 @@ func (c *CgroupsValidator) Name() string {
...
+				}
+				if len(fields)-i <= 3 {
+					return false, fmt.Errorf("expect 3 fields after `-` in mount info file %q", path)
```
neolit123, line 86:
```suggestion
					return false, fmt.Errorf("in mount info file %q, expected 3 fields after `-`, but got %d", path, xxxx)
```

also why can this happen?

validators/cgroup_validator_linux.go
```diff
@@ -44,28 +44,117 @@ func (c *CgroupsValidator) Name() string {
...
+				}
+				return strings.Contains(fields[i+2], "cgroup"), nil
+			}
+		}
+	}
+	return false, fmt.Errorf("path %q is not cgroupfs", path)
+}
+
+func parseMountsFile(path string) (string, error) {
```
neolit123, line 95: aren't we doing the same thing two times?

- checking a mount point path with isMountPointCgroupfs
- then reading the same file again with this function parseMountsFile

i think this can be done in one go.

here we don't care much about performance but it's just better code org.

validators/cgroup_validator_linux.go
pacoxu: This is following k/k code and I removed some logics that we did not depends on.

https://github.com/kubernetes/kubernetes/blob/07e73682b9a5259cacac978346ed4567588e4165/staging/src/k8s.io/mount-utils/mount_helper_unix.go#L102-L165

validators/cgroup_validator_linux.go
pacoxu: The current code is following the k/k. To make is simpler, I think we may split it using ` - `.
- There should be at least 3 fields after `-`.

validators/cgroup_validator_linux.go
pacoxu:
```
###  /proc/mounts is the same as  /proc/self/mounts
[root@paco ~]# cat /proc/mounts   | grep cgroup
cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0
none /run/cilium/cgroupv2 cgroup2 rw,seclabel,relatime 0 0
[root@paco ~]# cat /proc/self/mountinfo   | grep cgroup
28 22 0:25 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:4 - cgroup2 cgroup2 rw,seclabel
1499 27 0:25 / /run/cilium/cgroupv2 rw,relatime shared:573 - cgroup2 none rw,seclabel
```

Probably, we can just parce the ` /proc/mounts` for the first cgroup v2 path or first cgroup v1 path.

pacoxu:
I refactor the logic to just parse `/proc/mounts` to get the first cgroup path.
- and I add a simple test for `getUnifiedMountpoint( )`.

validators/cgroup_validator_test.go
```diff
@@ -93,3 +94,73 @@ func TestValidateCgroupSubsystem(t *testing.T) {
...
+			path, err := getUnifiedMountpoint(tmpFile.Name())
+
+			if test.expectedErr {
+				assert.Error(t, err, "%q: Expected error but got none", desc)
+			} else {
+				assert.NoError(t, err, "%q: Did not expect error but got one: %s", desc, err)
+			}
+
+			assert.Equal(t, test.expectedPath, path, "%q: Expected cgroup path mismatch", desc)
+		})
```
neolit123, line 164: we can remove the parameter `desc`  when doing assertion e.g.
`assert.Equal(t, test.expectedPath, path, "%q: Expected cgroup path mismatch")`

because the subtest `t.Run(desc, func(t *testing.T) {` already tells which test is currently running.

validators/cgroup_validator_test.go
```diff
@@ -93,3 +94,73 @@ func TestValidateCgroupSubsystem(t *testing.T) {
...
+		mountsFileContent string
+		expectedErr       bool
+		expectedPath      string
+	}{
+		"cgroup v2": {
+			mountsFileContent: "cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0",
+			expectedErr:       false,
+			expectedPath:      "/sys/fs/cgroup",
+		},
+		"cgroup v1": {
```
neolit123, line 109:
```suggestion
		"cgroups v2": {
			mountsFileContent: "cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0",
			expectedErr:       false,
			expectedPath:      "/sys/fs/cgroup",
		},
		"cgroups v1": {
```

validators/cgroup_validator_test.go
```diff
@@ -93,3 +94,73 @@ func TestValidateCgroupSubsystem(t *testing.T) {
...
+			mountsFileContent: "cgroup /sys/fs/cgroup cgroup rw,seclabel,nosuid,nodev,noexec,relatime 0 0",
+			expectedErr:       false,
+			expectedPath:      "/sys/fs/cgroup",
+		},
+		"empty file": {
+			mountsFileContent: "",
+			expectedErr:       true,
+			expectedPath:      "",
+		},
+		"none cgroup mounts": {
```
neolit123, line 119:
```suggestion
		"no cgroup mounts": {
```

validators/cgroup_validator_test.go
```diff
@@ -93,3 +94,73 @@ func TestValidateCgroupSubsystem(t *testing.T) {
...
+			expectedPath: "",
+		},
+		"multiple cgroups v1 and v2": {
+			mountsFileContent: `cgroup /sys/fs/cgroup/cpuset cgroup rw,nosuid,nodev,noexec,relatime,cpuset
+cgroup /sys/fs/cgroup/memory cgroup rw,nosuid,nodev,noexec,relatime,memory
+cgroup2 /sys/fs/cgroup/unified cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime`,
+			expectedErr:  false,
+			expectedPath: "/sys/fs/cgroup/unified",
+		},
+		"cgroup v1 only with multiple subsystems": {
```
neolit123, line 132:
```suggestion
		"cgroups v1 only with multiple subsystems": {
```

validators/cgroup_validator_test.go
```diff
@@ -93,3 +94,73 @@ func TestValidateCgroupSubsystem(t *testing.T) {
...
+			mountsFileContent: `cgroup /sys/fs/cgroup/cpuset cgroup rw,nosuid,nodev,noexec,relatime,cpuset
+cgroup /sys/fs/cgroup/memory cgroup rw,nosuid,nodev,noexec,relatime,memory`,
+			expectedErr:  false,
+			expectedPath: "/sys/fs/cgroup/cpuset", // First valid cgroup v1 path
```
neolit123, line 136:
```suggestion
			expectedPath: "/sys/fs/cgroup/cpuset", // First valid cgroups v1 path
```

validators/cgroup_validator_linux.go
```diff
@@ -45,27 +45,72 @@ func (c *CgroupsValidator) Name() string {
...
+	}
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	var cgroupV1MountPoint string
+	for scanner.Scan() {
+		line := scanner.Text()
+		if !strings.Contains(line, "cgroup") {
+			continue
+		}
+		// example fields: `cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0`
```
neolit123, line 66:
```suggestion
		// Example fields: `cgroup2 /sys/fs/cgroup cgroup2 rw,seclabel,nosuid,nodev,noexec,relatime 0 0`.
```

validators/cgroup_validator_linux.go
```diff
@@ -45,27 +45,72 @@ func (c *CgroupsValidator) Name() string {
...
+		fields := strings.Fields(line)
+		if len(fields) >= 3 {
+			switch fields[2] {
+			case "cgroup2":
+				// return the first cgroup v2 mount point directly
```
neolit123, line 71:
```suggestion
				// Return the first cgroups v2 mount point directly.
```

validators/cgroup_validator_linux.go
```diff
@@ -45,27 +45,72 @@ func (c *CgroupsValidator) Name() string {
...
+				return fields[1], nil
+			case "cgroup":
+				// set the first cgroup v1 mount point only,
+				// and continue the loop to find if there is a cgroup v2 mount point
```
neolit123, line 75:
```suggestion
				// Set the first cgroups v1 mount point only,
				// and continue the loop to find if there is a cgroup v2 mount point.
```

validators/cgroup_validator_linux.go
```diff
@@ -45,27 +45,72 @@ func (c *CgroupsValidator) Name() string {
...
+				if len(cgroupV1MountPoint) == 0 {
+					cgroupV1MountPoint = fields[1]
+				}
+			}
+		}
+	}
+	// return cgroup v1 mount point if no cgroup v2 mount point is found
```
neolit123, line 82:
```suggestion
	// Return cgroups v1 mount point if no cgroups v2 mount point is found.
```

validators/cgroup_validator_linux.go
```diff
@@ -45,27 +45,72 @@ func (c *CgroupsValidator) Name() string {
...
+			switch fields[2] {
+			case "cgroup2":
+				// return the first cgroup v2 mount point directly
+				return fields[1], nil
+			case "cgroup":
+				// set the first cgroup v1 mount point only,
+				// and continue the loop to find if there is a cgroup v2 mount point
+				if len(cgroupV1MountPoint) == 0 {
+					cgroupV1MountPoint = fields[1]
+				}
```
neolit123, line 78: is it ok to just do this?
```suggestion
				cgroupV1MountPoint = fields[1]
```

i.e. always update the var if more cgroups v1 mount points are found or should we only catch the first one?
(i don't know the format)

neolit123 (commented):
this is much better @pacoxu !
mostly requesting minor fixes for text / comments in multiple places.

and only one Q about the file parser.

validators/cgroup_validator_linux.go
pacoxu:
```
[root@10-29-14-249 ~]# cat /proc/mounts | grep cgroup
tmpfs /sys/fs/cgroup tmpfs ro,seclabel,nosuid,nodev,noexec,mode=755 0 0
cgroup /sys/fs/cgroup/systemd cgroup rw,seclabel,nosuid,nodev,noexec,relatime,xattr,release_agent=/usr/lib/systemd/systemd-cgroups-agent,name=systemd 0 0
cgroup /sys/fs/cgroup/blkio cgroup rw,seclabel,nosuid,nodev,noexec,relatime,blkio 0 0
cgroup /sys/fs/cgroup/hugetlb cgroup rw,seclabel,nosuid,nodev,noexec,relatime,hugetlb 0 0
cgroup /sys/fs/cgroup/perf_event cgroup rw,seclabel,nosuid,nodev,noexec,relatime,perf_event 0 0
cgroup /sys/fs/cgroup/memory cgroup rw,seclabel,nosuid,nodev,noexec,relatime,memory 0 0
cgroup /sys/fs/cgroup/net_cls,net_prio cgroup rw,seclabel,nosuid,nodev,noexec,relatime,net_prio,net_cls 0 0
cgroup /sys/fs/cgroup/cpu,cpuacct cgroup rw,seclabel,nosuid,nodev,noexec,relatime,cpuacct,cpu 0 0
cgroup /sys/fs/cgroup/devices cgroup rw,seclabel,nosuid,nodev,noexec,relatime,devices 0 0
cgroup /sys/fs/cgroup/pids cgroup rw,seclabel,nosuid,nodev,noexec,relatime,pids 0 0
cgroup /sys/fs/cgroup/cpuset cgroup rw,seclabel,nosuid,nodev,noexec,relatime,cpuset 0 0
cgroup /sys/fs/cgroup/freezer cgroup rw,seclabel,nosuid,nodev,noexec,relatime,freezer 0 0
```

For cgroup v1, this is an example. I prefer to get the first one.

validators/cgroup_validator_test.go
pacoxu: Good Catch.

neolit123 (commented):
/lgtm
/approve

should i create a new release after this?

pacoxu:
> should i create a new release after this?

Yes, as we have no todo items  for v1.32.

neolit123:
unrelated fix, please LGTM
https://github.com/kubernetes/system-validators/pull/42

will cut a release after it merges and send PR for k/k.

pacoxu:
/unhold
Thanks @KentaTada @neolit123 for your reviewing.

neolit123:
i will also rename the branch of this repo to main

PR #45, opened 2024-11-15
Drop use of deprecated `io/ioutil` package

tklauser:
The `io/ioutil` package was deprecated in Go 1.16[^1], which is the minimum required version per `go.mod`. Replace the uses of the package with the respective replacement functions.

[^1]: https://go.dev/doc/go1.16#ioutil


neolit123 (commented):
/lgtm
/approve
/ok-to-test

PR #46, opened 2024-11-15
Use `unix.Uname` instead of exec'ing `uname`

tklauser:
Instead of execing the `uname(1)` command, use the `uname(2)` syscall directly by means of the `unix.Uname` wrapper function to get the kernel release.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,4 +5,5 @@ go 1.16
 require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/stretchr/testify v1.3.0
+	golang.org/x/sys v0.27.0
 )
--- a/validators/os_validator_unix.go
+++ b/validators/os_validator_unix.go
@@ -21,8 +21,9 @@ package system
 
 import (
 	"fmt"
-	"os/exec"
 	"strings"
+
+	"golang.org/x/sys/unix"
 )
 
 var _ Validator = &OSValidator{}
@@ -39,11 +40,13 @@ func (o *OSValidator) Name() string {
 
 // Validate is part of the system.Validator interface.
 func (o *OSValidator) Validate(spec SysSpec) ([]error, []error) {
-	os, err := exec.Command("uname").CombinedOutput()
+	var utsname unix.Utsname
+	err := unix.Uname(&utsname)
 	if err != nil {
 		return nil, []error{fmt.Errorf("failed to get OS name: %w", err)}
 	}
-	if err = o.validateOS(strings.TrimSpace(string(os)), spec.OS); err != nil {
+	os := strings.TrimSpace(unix.ByteSliceToString(utsname.Sysname[:]))
+	if err = o.validateOS(os, spec.OS); err != nil {
 		return nil, []error{err}
 	}
 	return nil, nil
--- a/validators/package_validator_linux.go
+++ b/validators/package_validator_linux.go
@@ -159,15 +159,6 @@ func (validator *packageValidator) validate(packageSpecs []PackageSpec, manager 
 	return nil, errs
 }
 
-// getKernelRelease returns the kernel release of the local machine.
-func getKernelRelease() (string, error) {
-	output, err := exec.Command("uname", "-r").Output()
-	if err != nil {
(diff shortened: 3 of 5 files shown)


go.mod
```diff
@@ -5,4 +5,5 @@ go 1.16
 require (
 	github.com/blang/semver/v4 v4.0.0
 	github.com/stretchr/testify v1.3.0
+	golang.org/x/sys v0.27.0
```
neolit123, line 8: just to mention, this will not bring another dependency in kubernetes as it's already used there:
https://github.com/search?q=repo%3Akubernetes%2Fkubernetes%20golang.org%2Fx%2Fsys&type=code
and it's not an unwanted dependency.

https://github.com/kubernetes/kubernetes/blob/23e76432a9f2c602412d0d535fae4ac8d414d2ce/hack/unwanted-dependencies.json

validators/package_validator_linux.go
```diff
@@ -161,11 +162,12 @@ func (validator *packageValidator) validate(packageSpecs []PackageSpec, manager
...
 // getKernelRelease returns the kernel release of the local machine.
 func getKernelRelease() (string, error) {
-	output, err := exec.Command("uname", "-r").Output()
+	var utsname unix.Utsname
+	err := unix.Uname(&utsname)
 	if err != nil {
 		return "", fmt.Errorf("failed to get kernel release: %w", err)
 	}
-	return strings.TrimSpace(string(output)), nil
+	return strings.TrimSpace(unix.ByteSliceToString(utsname.Release[:])), nil
```
neolit123, line 170: this produces exactly the same output as `uname -r`. i confirmed locally.
but there are no real benefits for it, unless some distros don't have `uname` installed.
@tklauser have you seen such cases?

neolit123 (commented):
i think this is fine, just trying to understand the reason for the change.
/cc @pacoxu

validators/package_validator_linux.go
tklauser: @neolit123 I haven't seen such a case in the real world but I could imagine minimal locked-down distributions not shipping or blocking access to `uname`. Plus it's a single `uname` syscall instead of `fork`/`exec` which I guess doesn't really matter here but is still nice if it can be avoided. So it felt like an easy change to to, especially `x/sys` is already a dependency in kubernetes.

But feel free to disregard if this feels like churn.

neolit123: no, it seems good to me.
/approve

pacoxu: https://github.com/kubernetes/system-validators/blob/85af5a8471fc6a9ce319eb637654432992f52c6c/validators/os_validator_unix.go#L42 

https://github.com/kubernetes/system-validators/blob/85af5a8471fc6a9ce319eb637654432992f52c6c/validators/types_unix.go#L99

Should these two be updated as well if so?

tklauser: @pacoxu thanks, I've updated these too.

validators/os_validator_unix.go
```diff
@@ -39,11 +40,13 @@ func (o *OSValidator) Name() string {
...
 // Validate is part of the system.Validator interface.
 func (o *OSValidator) Validate(spec SysSpec) ([]error, []error) {
-	os, err := exec.Command("uname").CombinedOutput()
+	var utsname unix.Utsname
+	err := unix.Uname(&utsname)
 	if err != nil {
 		return nil, []error{fmt.Errorf("failed to get OS name: %w", err)}
 	}
-	if err = o.validateOS(strings.TrimSpace(string(os)), spec.OS); err != nil {
+	os := strings.TrimSpace(unix.ByteSliceToString(utsname.Sysname[:]))
```
neolit123, line 48: here it's possible to call getKernelRelease as well?

tklauser: `getKernelRelase` returns `utsname.Release` (i.e. the kernel version) while this needs `uts.Sysname` (i.e. the kernel name), so I don't think we can use `getKernelRelase` here.

neolit123 (commented):
/lgtm

this code will be included in the next release of the lib.
thanks @tklauser

Issue #51, opened 2025-09-12
[need-info] Does kubeadm require CONFIG_CGROUP_CPUACCT/DEVICE/FREEZER even with cgroups v2?


Priyankasaggu11929:
Is it correct understanding that the following kernel config options (from the cgroups v1) are always need to be set/enabled by default on the node host, regardless of whether the node is configured to use cgroups v1 or v2?

```
CONFIG_CGROUP_CPUACCT: not set
CONFIG_CGROUP_DEVICE: not set
CONFIG_CGROUP_FREEZER: not set
```

If so, is this due to these kernel configs hardcoded in the following 2 places: 

* [https://github.com/kubernetes/system-validators/blob/c8b796a384a0c0a69eb7f00ba61cad931780fc3a/validators/types\_unix.go#L46-L48](https://github.com/kubernetes/system-validators/blob/c8b796a384a0c0a69eb7f00ba61cad931780fc3a/validators/types_unix.go#L46-L48)
* [https://github.com/kubernetes/system-validators/blob/c8b796a384a0c0a69eb7f00ba61cad931780fc3a/validators/types\_unix.go#L71](https://github.com/kubernetes/system-validators/blob/c8b796a384a0c0a69eb7f00ba61cad931780fc3a/validators/types_unix.go#L71)

(For context, this is related to the OpenSUSE MicroOS community ticket – [https://bugzilla.suse.com/show\_bug.cgi?id=1249454](https://bugzilla.suse.com/show_bug.cgi?id=1249454))

Thank you so much, in advance for your help!

cc: @pacoxu

Priyankasaggu11929:
Thank you for the help, @neolit123 @pacoxu 🙏

Priyankasaggu11929:
Thank you so much @neolit123 @pacoxu, for the help here 🙏

PR #53, opened 2025-09-16
Remove cgroup v1/v2 only Kernel Config checks

pacoxu:
https://man7.org/linux/man-pages/man7/cgroups.7.html (search `in Linux 5.2 and
              earlier).`)

https://github.com/kubernetes/system-validators/issues/51#issuecomment-3285257204

To make it cherry-pickable, I opened this PR.

### DIFF
--- a/validators/types_unix.go
+++ b/validators/types_unix.go
@@ -42,12 +42,6 @@ var DefaultSysSpec = SysSpec{
 			{Name: "IPC_NS"},
 			{Name: "UTS_NS"},
 			{Name: "CGROUPS"},
-			{Name: "CGROUP_BPF"},     // cgroups v2
-			{Name: "CGROUP_CPUACCT"}, // cgroups v1 cpuacct
-			{Name: "CGROUP_DEVICE"},
-			{Name: "CGROUP_FREEZER"}, // cgroups v1 freezer
-			{Name: "CGROUP_PIDS"},
-			{Name: "CGROUP_SCHED"}, // cgroups v1 & v2 cpu
 			{Name: "CPUSETS"},
 			{Name: "MEMCG"},
 			{Name: "INET"},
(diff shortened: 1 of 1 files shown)


neolit123 (commented):
if you are going with removing the options, best to rename the commit and pr name.

pacoxu:
> > if you are going with removing the options, best to rename the commit and pr name.
> 
> This was addressed.
> 
> ~If we backport this, it means users can pass the check without CGROUP_BPF for cgroup v2; and can pass without those CPUACCT/DEVICE/FREEZER for cgroup v1. Is that a bigger issue?~

Sorry. My fault. I forgot the cgroup controller checks. 

It would be OK. 

/unhold

neolit123:
> I created a CR #53 here.

i'm getting a bit confused.

i think we need to do this:
- have a single PR / commit that removes all v1 and v2 options, leaves only `{Name: "CGROUPS"},`
- we merge it main here and backport it to SV versions, then backport it to k/k
- it makes sure users of v1 or v2 don't get any errors if they are missing options
- the other bigger PR can split the config options, we release a new MINOR here and add it to the latest k/k release

does that plan make sense?

pacoxu:
> > I created a CR #53 here.
> 
> i'm getting a bit confused.
> 
> i think we need to do this:
> 
> * have a single PR / commit that removes all v1 and v2 options, leaves only `{Name: "CGROUPS"},`
> * we merge it in main here and backport it to SV versions, then backport it to k/k
> * it makes sure users of v1 or v2 don't get any errors if they are missing options
> * the other bigger PR can split the config options, we release a new MINOR here and add it to the latest k/k release
> 
> does that plan make sense?

Totally agree.

pacoxu:
Updated this PR and the CR #54

pacoxu:
> 1.9.2 https://github.com/kubernetes/system-validators/releases/tag/v1.9.2
> 
> do we not need this in 1.10 as well?

We may release a v1.10 after #52 is merged. I will rebase with this one.

PR #63, opened 2026-05-12
add nftables system validator

aojea:
First check if the kernel modules is loaded in /sys/module/nf_tables (does not require root).

If the file does not exist, then try to list all nftables tables via netlink (it does require root priviledges).


pacoxu:
> For Kubernetes 1.36, the [nftables mode](https://kubernetes.io/docs/reference/networking/virtual-ips/#proxy-mode-nftables) of kube-proxy requires version 1.0.1 or later of the nft command-line tool, as well as kernel 5.13 or later.

https://kubernetes.io/docs/reference/node/kernel-version-requirements/#kube-proxy-nftables-proxy-mode

Do we need to add the kernel version and nft version check here?

aojea:
> Do we need to add the kernel version and nft version check here?

I think is already covered by 

https://github.com/kubernetes/system-validators/blob/9040bc28d5fabd3fcfe4e9c74e4b294a4dbf5e67/validators/types_unix.go#L32-L35

but please keep me honest

danwinship:
> > Do we need to add the kernel version and nft version check here?
> 
> I think is already covered by
> 
> https://github.com/kubernetes/system-validators/blob/9040bc28d5fabd3fcfe4e9c74e4b294a4dbf5e67/validators/types_unix.go#L32-L35

5.4 and 5.10 are too old.

(Well, in fact, we _really_ only need kernel 5.3, but kube-proxy will refuse to run with anything older than 5.13 by default, because it assumes if you have an older kernel than that, then you also have an old `nft` binary in the host filesystem, and we don't want to run kube-proxy nftables on systems where the host system's `nft` binary is too old, per https://github.com/kubernetes/kubernetes/issues/122743#issuecomment-1893922424.)

pacoxu:
> > > Do we need to add the kernel version and nft version check here?
> > 
> > 
> > I think is already covered by
> > https://github.com/kubernetes/system-validators/blob/9040bc28d5fabd3fcfe4e9c74e4b294a4dbf5e67/validators/types_unix.go#L32-L35
> 
> 5.4 and 5.10 are too old.
> 
> (Well, in fact, we _really_ only need kernel 5.3, but kube-proxy will refuse to run with anything older than 5.13 by default, because it assumes if you have an older kernel than that, then you also have an old `nft` binary in the host filesystem, and we don't want to run kube-proxy nftables on systems where the host system's `nft` binary is too old, per [kubernetes/kubernetes#122743 (comment)](https://github.com/kubernetes/kubernetes/issues/122743#issuecomment-1893922424).)

In this case, a warning about older kernel version than 5.13 for `nft` would be needed. Or check the `nft` binary directly?

aojea:
> In this case, a warning about older kernel version than 5.13 for `nft` would be needed. Or check the `nft` binary directly?

@danwinship didn't we assume the nft binary will be part of the kube-proxy image?

danwinship:
kube-proxy's `nft` will be part of the kube-proxy image, but as per the discussion in #122743, we care about the host's `nft` binary too.

> Or check the `nft` binary directly?

kube-proxy itself will still check for kernel 5.13, so unless you're planning to run it with `KUBE_PROXY_NFTABLES_SKIP_KERNEL_VERSION_CHECK=true`, then the _actual_ host `nft` version doesn't matter if the kernel is too old.

pacoxu:
> kube-proxy's `nft` will be part of the kube-proxy image, but as per the discussion in #122743, we care about the host's `nft` binary too.
> 
> > Or check the `nft` binary directly?
> 
> kube-proxy itself will still check for kernel 5.13, so unless you're planning to run it with `KUBE_PROXY_NFTABLES_SKIP_KERNEL_VERSION_CHECK=true`, then the _actual_ host `nft` version doesn't matter if the kernel is too old.

Got it.


I saw the discussion https://github.com/kubernetes/enhancements/pull/5344#discussion_r3215049219 about whether we should add the error message here if iptables is still the default and not deprecated. 

Besides, @neolit123 add a warning in https://github.com/kubernetes/kubernetes/pull/139067.

PR #64, opened 2026-09-18
Add tests for the top level Validate and ValidateSpec dispatcher

magic-peach:
Adds unit tests for validators.Validate and validators.ValidateSpec, the top-level dispatcher functions in validators/validators.go, which previously had no direct test coverage at all (only the individual per-validator packages were tested).

Validate is tested with mock Validator implementations covering an empty validator list, all-clean validators, and validators returning only warnings, only errors, or both, and asserts the returned slices are properly aggregated and ordered. ValidateSpec is tested across several containerRuntime values to confirm it returns without panicking; an empty SysSpec has a nil DockerSpec so this does not exec a real docker binary.

Testing: `go test ./validators/... -v` and `go test ./... -cover` both pass, package coverage went from 47.6% to 67.5%. `go vet` and `gofmt -l` are clean.


neolit123 (commented):
/lgtm
/approve

thanks for the PR.

